Issue #1, opened 2017-10-11
GlyphsInfo badly migrated


thlinard:
https://github.com/googlefonts/tools/tree/master/util/GlyphsInfo returns a 404 error.

davelab6:
@graphicore please could you fix this? :)

graphicore:
related graphicore/googleFontsTools#1

rsheeter:
Fixed by #2. Thanks for the report.

davelab6:
Despite merging #2, https://github.com/googlefonts/tools/tree/master/util/GlyphsInfo still 404s for me :)

graphicore:
Yeah, if navigate to it via the [github interface](https://github.com/googlefonts/tools/tree/master/util) you'll end up here: https://github.com/schriftgestalt/GlyphsInfo/tree/e33ccf3515cc5b8005a3a50b4163663623649d20 I think that redirection was broken before.

davelab6:
Hah ok, thanks for confirming

PR #2, opened 2017-10-11
Add .gitmodules; Update util/GlyphsInfo to current master. See #1

graphicore:


graphicore:
So, who's responsible for merging into this? :-)

PR #23, opened 2017-11-03
Cyrillic Filter Lists housekeeping

alexeiva:
Removed optional .sc glyphs from Cyrillic Plus filter list. They are reduplicated in a separate list. 

This was a mistake, their existence in this list created confusion.


thlinard:
Interesting. @alexeiva do you think the same treatment should be applied to the Greek filter lists?

alexeiva:
Yes, I think so. Small Caps are unencoded glyphs, and GF glyphs sets have a specified set of encoded glyphs, and a recommended set of unencoded glyphs.

thlinard:
OK, I don't have time now, but I'll look at it as soon as possible.

alexeiva:
Just realized [Latin Expert](https://github.com/googlefonts/tools/tree/master/encodings/GF%20Glyph%20Sets#latin) explicitly supports Small Caps.

I think Small Cap support for Cyrillic should simply depend on whether they are supported in Latin. So I'll make an optional small cap list, and update the docs.

Same for Greek?

thlinard:
I totally agree. If the fonts support SC for Latin, SC support is a legitimate expectation for Cyrillic and Greek.

It's an issue that I observed with the new version of EB Garamond: a good coverage of characters [has been reached](https://github.com/octaviopardo/EBGaramond12/issues/6), but for example the support of the small caps in Greek is incomplete: some characters will pass in small caps, others not. I'm glad that version 1.0 is out, it was long overdue, but for future versions, a tool to check the support of glyphs, a sort of https://graphicore.github.io/mdlFontSpecimen/html/language-coverage-details.html for glyphs, would be good.

PR #38, opened 2017-11-27
GF Glyph Sets/Vietnamese/README.md Updated

davelab6:
@alexeiva please review :)

### DIFF
--- a/encodings/GF Glyph Sets/Vietnamese/README.md
+++ b/encodings/GF Glyph Sets/Vietnamese/README.md
@@ -3,6 +3,7 @@
 Here is how to create localized Vietnamese glyphs, different from default Latin ones, in Glyphs App.
 
 First, let's create glyphs for localized Vietnamese accents. 
+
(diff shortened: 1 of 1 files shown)


felipesanches:
@davelab6 opened the PR asking for review and @alexeiva approvedthe proposed changes, but nobody merged this. For some reason I cannot merge this myself. In the past I used to have the proper write access to do so. Not sure what happened since.

@davelab6 if you want me to merge PRs here, then we'd need to ensure I have the proper write access rights to do so.

felipesanches:
I vaguely recall that the Google Fonts team mirrors this repo internally, and that this could be a reason why they'd be willing to lead the process of reviewing PRs (or something like that) is that correct?

I'm asking this simply because I'm concerned that we have a couple of open PRs in this repo that were not merged after several months.

davelab6:
@felipesanches thanks for the ping. This change is to a README so would not effect the GF staff eng team

PR #58, opened 2018-06-06
Add check-vf-avar

m4rc1e:
This script will generate a html doc which compares a family of static fonts against a family of variable fonts instances.

It works by overlaying the variable font instances on top of the static fonts. This should help us spot anything that may be off with a variable font's avar table.

screen shot 2018-06-06 at 11 54 41
*Variable font's avar table matches the static fonts*

screen shot 2018-06-06 at 11 56 17
*Variable font's avar doesn't match the static fonts*

It is not a beautiful script, but it will do until we get this functionality integrated into GF Regression.

### DIFF
--- a/bin/gftools-check-vf-avar.py
+++ b/bin/gftools-check-vf-avar.py
@@ -1,0 +1,216 @@
+"""
+Generate a html overlay doc which compares a family of static fonts against
+a family of vf font instances.
+
+If the variable font instances do not match the static fonts perfectly,
+it usually means the avar table needs adjusting
+https://docs.microsoft.com/en-us/typography/opentype/spec/avar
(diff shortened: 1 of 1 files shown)


m4rc1e:
Let me update this.

I'd prefer the styles to be sorted by weight order.

m4rc1e:
Alright, output is now sorted by weight

screen shot 2018-06-06 at 12 28 24

davelab6:
This is a nice and simple solution, thanks Marc!

m4rc1e:
If this avar matching gets approved, we should have an FB test for this

Its not hard to work out what the recommended values should be, it's just normalisation.

PR #79, opened 2018-08-28
Add new tool to check for expected caret count

eliheuer:
This tool will check a font and determine the expected caret count.
Then it will find the actual caret count and compare the expected
count against the actual count for each glyph.

The user will be warned about discrepancies.


davelab6:
Shouldn't this be a fontbakery check?

PR #81, opened 2018-08-31
Update build-ofl script for Python3 compatibility

### LINKED ISSUE #68: Python 2-3 compatibility
I did some preliminary tests and it seems like making all of gftools python 2-3 compatible should be a simple task. 

This would make my workflow much easier since I like to work only with python3 venvs. 

It would also help make the install instructions in the README.md easier to follow.

eliheuer:
This update will allow the script `build-ofl` to work with both
Python2 and Python3

Fixes #68 for this script


m4rc1e:
Thanks for this!

PR #83, opened 2018-09-17
fix-vf-meta.py: Fix script so fonts can work in MS Word

m4rc1e:
- Downgraded STAT table
- fix psname

Fonts simply didn't work in Word because Ms Word doesn't yet support the latest STAT table version. I need to get in touch with MS about this.


davelab6:
Oh boy.

m4rc1e:
Non Ribbi styles are now selectable as well.

screen shot 2018-09-17 at 15 45 41

Unfortunately the script still only works for VFs which only have a weight axis. I spoke to Garret and they're in a similar position.

davelab6:
@m4rc1e ready to merge?

m4rc1e:
Yeah, lets merge it.

It still only support single axis fonts (same state as GF). I'll keep updating this.

PR #101, opened 2018-12-06
WIP gftools qa: added

m4rc1e:
Our GF QA uses three tools, fontbakery, fontdiffenator and diffbrowsers. This script simply orchestrates them. For making comparisons, it can fetch fonts from Google Fonts or a user's system.

The aim of this script is to reduce the whole QA process to a single click. I predominately rely on the following two commands:

**QAing new families**
`gftools qa [fonts] --fontbakery --plot-glyphs --browser-previews`

**QAing updated families which already exist on GF**
`gftools qa [fonts] -gf --fontbakery --plot-glyphs --diffbrowsers`

I found myself writing something similar whilst working on CI stuff. I think it's better to have a standalone script so that both Travis and individuals can use it. I also include it in my fontmake build scripts. Automated builds and tests are a true life saver.

For people who don't know our tools, this script does the following.

- Font linting: Does the metadata in each font pass our fontbakery spec
- Font regressions: How do the fonts compare against another set of fonts
- Font hinting: How do the fonts look on modern browsers against another set of fonts
- Glyph plots: Show me every glyph

For font hinting, you'll need a browserstack account. Font regressions relies on FreeType, HB and Cairo.

 need to finish off https://github.com/googlefonts/fontdiffenator/pull/35 before this can be merged.

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -17,29 +17,75 @@ from fontTools import ttLib
 import requests
 from io import BytesIO
 from zipfile import ZipFile
+import sys
+import os
+import shutil
+if sys.version_info[0] == 3:
+    from configparser import ConfigParser
+else:
+    from ConfigParser import ConfigParser
 
 # =====================================
 # HELPER FUNCTIONS
 
-def download_family_from_Google_Fonts(family_name):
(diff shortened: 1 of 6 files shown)


m4rc1e:
This can be merged. There's probably a few edge cases I've missed. I'll keep working on it.

Issue #92, opened 2018-10-25
gftools-add-font.py: Needs documentation and usability improvements


eliheuer:
Myself and others are having trouble using `add-font.py`, it seems like this tool could use some refactoring and documentation. Marc has said the following to me regarding this tools: 

> frankly that should be sorted out before py3
> and it should be done with the GF eng team on board 

## Issue One
One issue I run into is needing gflags:

![screen shot 2018-10-19 at 12 04 39 pm](https://user-images.githubusercontent.com/5162664/47503265-ace09380-d837-11e8-942f-9e13146812df.png)

I have found that the python version works, `pip2 install python-gflags`

## Issue Two
I believe some of the paths are relative, this causes problems because the tool must be run from a specific directory, see below: 

![screen shot 2018-10-25 at 9 29 25 am](https://user-images.githubusercontent.com/5162664/47503593-8111dd80-d838-11e8-9074-778ba0c2d8ef.png)

eliheuer:
I made a short video showing how to use this tool at this point in time: https://youtu.be/BTMPhh2737A

davelab6:
> Issue One

Since `pip2 install python-gflags`, why not include `python-gflags` in `requirements.txt` and be done?

Perhaps because https://github.com/google/python-gflags is Py2 only? However, its README says that its been abandoned and progress happens in "abseil", and that is Py23 compatible per https://github.com/abseil/abseil-py/blob/181c8b334b87057d5dd32861f9708fb8ee66189e/setup.py#L33

So, include `absl-py` in `requirements.txt` and be done? :)

> Issue Two

If some of the paths are relative, what needs to be done to make the paths work from anywhere? :)

If you (@eliheuer) can make a PR that fixes this for you, then @nyshadhr9 or someone on the team can review that PR and let you know if your changes can be accepted into the google-internal "upstream" copy of the tool :)

eliheuer:
Sounds good, I can try a PR.

This is an important tool and probably one of the first ones new users will try, so it should be easy to use from both py2 and py3.

eliheuer:
We also need to add some documentation around using this with variable fonts: 

```
$ python2 bin/gftools-add-font.py ~/Google/fonts/ofl/kreon

Traceback (most recent call last):
...
  File "/Users/Air/Demo/gftools/Lib/gftools/util/google_fonts.py", line 446, in StyleWeight
    return ('normal', _KNOWN_WEIGHTS[styleweight])
KeyError: 'VF'
```

[...]

PR #105, opened 2019-01-04
Fix add-font.py relative path to encodings issue #92

eliheuer:
This PR fixes the relative path to encodings problem that was discussed in issue #92, I have a lengthy [post](https://github.com/googlefonts/gftools/issues/92#issuecomment-451381849) there discussing this solution.


m4rc1e:
Unfortunately this won't work if a user pip installs gftools.

When I check the `enc_path` when using a pip install of this pr, I get the following path.
`/Users/marcfoley/Desktop/venv/encodings/`

My venv dir doesn't contain the encodings dir.

When I attempt to use the script on an ofl dir. Only the menu subset is returned.

anthrotype:
relying on sys.path like that to get some data directory path is not a good idea. You're making lots a assumption about how that python enviornment is set up.
If those encodings are meant to be used by your script, and available not just when running it from the source checkout but also after installing normally it via pip, I'd recommend that you add it to the `package_data` in setup.py, so they will be installed alogside gftools package. Then from the script you can locate the directory relative to `os.path.dirname(gftools.__file__)`, or use methods from pkg_resurces or pkgutil

m4rc1e:
@eliheuer do you need help with this?

m4rc1e:
I am ok with my commit being included

m4rc1e:
@eliheuer I've updated this pr.

I've done the following:
- Moved `encodings/` to `Lib/gftools/encodings/`
- include `encodings/` in `setup.py.package_data`
- Update `nam_dir` flag in `Lib/gftools/util/googlefonts.py` to load the `package_data` by default
 
Please repull and test it.

This pr won't affect the engineers if they're specifying a nam_dir from the cli either.

m4rc1e:
@eliheuer if you're happy with my changes, you may want to tidyup the last commit a bit (whitespace etc). Do that then squash and force push.

m4rc1e:
Thanks Cosimo! to be on the safe side I added setuptools to the setup.py.

@eli please feel free to review this. If it work for you, merge away.

eliheuer:
Hmm... why is there a CLA issue with e58fa6f ?

eliheuer:
Ok, I tested this and it works for enough to merge. It still needs some help parsing VF file names, but that can be another PR.

PR #114, opened 2019-01-29
WIP: gftools-fix-isfixedpitch.py added.

m4rc1e:
Added simple script to set the post tables' isfixedpitch attrib to 1 if monospace else 0.

@thundernixon this should be a good starting point. I'd look into what other settings should exist for monospace fonts.

### DIFF
--- a/bin/gftools-fix-isfixedpitch.py
+++ b/bin/gftools-fix-isfixedpitch.py
@@ -1,4 +1,39 @@
-"""Fix a collection of fonts isfixedpitch flag"""
+#!/usr/bin/env python
+
+# Copyright 2013,2016 The Font Bakery Authors.
+# Copyright 2017 The Google Font Tools Authors
+#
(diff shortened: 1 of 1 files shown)


bin/gftools-fix-isfixedpitch.py
```diff
@@ -0,0 +1,38 @@
+"""Fix a collection of fonts isfixedpitch flag"""
+from fontTools.ttLib import TTFont
+import argparse
+
+
+def fix_isFixedPitch(ttfont):
+
+    same_width = set([])
```
moyogo: Why not just `same_width = set()`?

m4rc1e: Sorry this was a demo script

m4rc1e: I guess we'll improve it in time. Was done over a google hangout.

PR #116, opened 2019-02-04
WIP: check-sandbox-family added

m4rc1e:
This script enables us to swap any website's GF hosted fonts with the same fonts hosted on our sandbox . Once we've finished QAing a family and the family is sandboxed, this should be the final check imo.

Here's a [production vs sandbox gif](https://user-images.githubusercontent.com/7525512/52221219-9d420380-2898-11e9-963e-dbf177316963.gif) for IBM Plex Sans on https://www.designworkplan.com/

@chrissimpkins This approach may interest you for your projects.

This is still WIP. Needs px diffing etc

### DIFF
--- a/bin/gftools-check-sandbox-family.py
+++ b/bin/gftools-check-sandbox-family.py
@@ -1,0 +1,109 @@
+#!/usr/bin/env python
+"""
+Make a diff/gif image of a site by swapping the GF production families
+with the same families hosted on the sandbox server.
+
+
+Usage:
+gftools check-sandbox-family https://www.somesite.com
+"""
+import os
+import sys
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options 
+from PIL import Image, ImageDraw, ImageFont
+import argparse
+from urllib.parse import urlsplit
+from io import BytesIO
+import time
+
+
+SWAP_FONT_JS = """
+links = document.getElementsByTagName('link');
+for (i in links) {
+    if (typeof(links[i].href) != 'undefined') {
+         if (links[i].href.indexOf("fonts.googleapis.com") > -1)             
+             links[i].href = links[i].href.replace("fonts.googleapis", "fonts.sandbox.google");
+    }
+}
+"""
+WIDTH = 1024
+
+
+def get_font_for_os():
+    if sys.platform.startswith("linux"):
+        return os.path.join(
+                "usr", "share", "font", "truetype", "noto"
+                "NotoMono-Regular.ttf")
+    elif sys.platform.startswith("darwin"):
+        return os.path.join("Library", "Fonts", "Arial.ttf")
+    elif sys.platform.startswith("win"):
+        return os.path.join("c:", "\\", "Windows", "Fonts", "arial.ttf")
+    else:
+        raise NotImplementedError("Please use OSX, Ubuntu or Win")
+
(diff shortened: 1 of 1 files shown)


bin/gftools-check-sandbox-family.py
```diff
@@ -0,0 +1,85 @@
...
+        driver.set_window_size(WIDTH, required_height)
+
+        try:
+            families_in_use = driver.find_elements_by_xpath(
+                    '//link[contains(@href, "fonts.googleapis.com/css")]'
+            )
+            for family in families_in_use:
+                print("Changing GF url %s to %s" % (
+                    family.get_attribute("href"), family.get_attribute("href").replace(
+                        "fonts.googleapis.com", "fonts.sandbox.google.com")
```
chrissimpkins: `fonts.sandbox.google.com` can be replaced with any GET request accessible URL to a CSS file?

chrissimpkins: Ah! I see it up here:

https://github.com/googlefonts/gftools/pull/116/files#diff-2331a623dfd3e84252439986a266425cR23

I will take a closer look at this.  Thanks Marc!

bin/gftools-check-sandbox-family.py
```diff
@@ -0,0 +1,85 @@
+"""Make a diff/gif image of a site by swapping the GF production families
+with same families hosted on the sandbox server.
+
+
+Usage:
+gftools check-sandbox-family https://www.somesite.com
+"""
+import os
+from selenium import webdriver
+from selenium.webdriver.chrome.options import Options
```
davelab6: Is this added to the root project `requirements.txt`?

bin/gftools-check-sandbox-family.py
davelab6: Making a more generic swapper is fine, but I think we need a script specifically for our sandbox.

bin/gftools-check-sandbox-family.py
```diff
@@ -0,0 +1,85 @@
...
+        before_img = driver.get_screenshot_as_png()
+        js = JS
+        driver.execute_script(js)
+        time.sleep(5)
+        after_img = driver.get_screenshot_as_png()
+
+        gif_path = urlsplit(args.url).netloc + ".gif"
+
+        with Image.open(BytesIO(before_img)) as before, Image.open(BytesIO(after_img)) as after:
+            font = ImageFont.truetype("/Library/Fonts/Arial.ttf", 32)
```
davelab6: This path is macOS specific, and AFAIK the rest of the script is not; it would be better to bundle a Roboto-Regular.ttf subset to ASCII and use that, I think.

davelab6:
I dropped a few comments but over LGTM to merge

bin/gftools-check-sandbox-family.py
m4rc1e: Not yet

bin/gftools-check-sandbox-family.py
m4rc1e: I'll use os.path so it's OS agnostic. Thanks.

m4rc1e:
@davelab6 Thanks, I've tidied up the script. This is now good to merge.

davelab6:
Sorry my delay.

PR #125, opened 2019-02-25
lower minimum percentage needed to apply "ext" subset

thundernixon:
This is a patch for https://github.com/google/fonts/issues/187.

It may or may not be the best solution for the problem, but as I understand it, we want to add `ext` subsets into metadata if fonts contain even a single glyph of a given `ext` subset. By setting the default `min_pct_ext` to `0.01`, this mean that if 1 glyph of up to 1000 is present from an `ext` encoding file, it will apply that subset label. Because Cyrillic has the most glyphs in an `ext` encoding file at 333 glyphs, this gives some headroom. 

It doesn't apply `ext` subsets to all fonts, so it seems to be doing the job.

There are still further edits to make to `add-font`, but this one should correct this specific issue in future uses.

### DIFF
--- a/bin/gftools-add-font.py
+++ b/bin/gftools-add-font.py
@@ -62,9 +62,10 @@ FLAGS = flags.FLAGS
 flags.DEFINE_integer('min_pct', 50,
                      'What percentage of subset codepoints have to be supported'
                      ' for a non-ext subset.')
-flags.DEFINE_integer('min_pct_ext', 10,
-                     'What percentage of subset codepoints have to be supported'
-                     ' for a -ext subset.')
+# if a single glyph from the 81 glyphs in *-ext_unique-glyphs.nam file is present, the font can have the "ext" subset
+flags.DEFINE_float('min_pct_ext', 0.01,
+                   'What percentage of subset codepoints have to be supported'
+                   ' for a -ext subset.')
 
 
 def _FileFamilyStyleWeights(fontdir):


m4rc1e:
> but as I understand it, we want to add `ext` subsets into metadata if fonts contain even a single glyph of a given `ext` subset.

Just a single glyph? I'd expect much more than that. What's wrong with the original 10% val?

thundernixon:
From issue google/fonts#187

> @thundernixon
> Are you saying that the latin-ext subset should be enabled if there are any characters inside that range? And I'm guessing we would do this by including this in the METADATA.pb file?
 > 
> @davelab6
> Yes, yes. There is existing software to identify if a font supports the latin-ext range, I think the add-font script.
 
Without the subset label, people can't access a glyph that is outside of the subset.

Of course, now that there are Unicode ranges in modern browsers, this is somewhat less urgent, but still worth fixing.

m4rc1e:
Imo adding just a single glyph to enable a codepage kind of defeats the point of them.

The Latin-ext codepage has 813 glyphs. The current implementation expects that at least 81 glyphs exist before it gets enabled. One can assume that those 83 glyphs will add support for at least another language. The approach you're suggesting will trigger this codepage if someone just adds a currency symbol such as a rupee.

thlinard:
@m4rc1e : 10 % is largely insufficient for greek-ext. When you want to set texts in Polytonic Greek (the only "language" used by greek-ext), you need the full range of greek-ext, not 10%.

thundernixon:
Okay, yeah I'm convinced that this isn't the right solution, but rather that @thlinard's recent suggestion on GF issue 187 would be a better endpoint: separating the logic for subsetting from the logic of language support.

davelab6:
> The Latin-ext codepage has 813 glyphs. The current implementation expects that at least 81 glyphs exist before it gets enabled. One can assume that those 83 glyphs will add support for at least another language. The approach you're suggesting will trigger this codepage if someone just adds a currency symbol such as a rupee.

This isn't a codepage, its a subset.

If someone just has a currency symbol such as a rupee, then its not available unless latin-ext is enabled. 

Therefore I suggest different % values for different subsets; 100% for greek-ext is fine, and 0.1% for latin-ext is fine.

Issue #122, opened 2019-02-13
add-font adds spaces to ABeeZee font name


thundernixon:
I'm using `add-fonts` to generate new metadata for all families in google/fonts, in order to make sure we don't exclude latin-ext glyphs where they exist (see https://github.com/google/fonts/issues/187). This has led me to find a number of issues with `add-font`. I'll document them separately, so that we can hopefully correct things that are real issues, and I can be better informed if some of them aren't real issues.

---

If `add-font` is run on ABeeZee, the `METADATA.pb` file ends up with 

```
name: "A Bee Zee"
```

However, in the `name` table of ABeeZee, `nameID` 1 is 

```
    <namerecord nameID="1" platformID="1" platEncID="0" langID="0x0" unicode="True">
      ABeeZee
    </namerecord>
```

(without spaces)

So, we may need to add an exception for ABeeZee, unless this is really the intended behavior.

felipesanches:
fontbakery does have exceptions for this kind of thing. Maybe we can copy the exceptions into gftools add-font as well

felipesanches:
https://github.com/googlefonts/fontbakery/blob/d2cb61bf4c1f97b03f437a0f2da3d454c8c9b6c9/Lib/fontbakery/specifications/googlefonts.py#L2838-L2851

thundernixon:
That's pretty much what I had in mind! I thought it actually did exist in add-font, but that must have been in another tool.

m4rc1e:
I don't like this camelcase splitting function. I also use it in GFR. Couldn't we get the family name in the following manner?

```
nametable = font['name']

family_name = None
typographic_name = nametable.getName(16, 3, 1, 1033)
if typographic_name:
    family_name = typographic_name.toUnicode()
else:
    family_name = nametable.getName(1, 3, 1, 1033).toUnicode()
```

m4rc1e:
The above method does depend on the fonts having good name tables though? so maybe the filename may be a better indicator of the family name. Anyone willing to test this across our collection?

thundernixon:
I think I will probably need to check name changes anyway, so I'd be happy to try this. 

What is nameID `1033`? I don't see it in the MS spec, and I don't recall coming across it before.

m4rc1e:
it's the langID of the record.

PR #126, opened 2019-02-25
add-font: get actual font name for metadata rather than splitting file name

thundernixon:
This aims to resolve https://github.com/googlefonts/gftools/issues/122

I've added Marc's suggested code to get the font nameID 16, else nameID 1, and this results in a more-accurate result for font names in metadata generated by `gftools-add-font.py`.

I have a few questions before I'll feel confident in this merge:

- **Is there a cleaner way to add this functionality?** E.g. should the new `_getFontFamilyName` function be placed somewhere else, such as in `util/google_fonts.py`?

- **Will this work for VFs?** Might it accidentally include a style name, because nameID 1 will likely include the default style name? (Also, in the context of this update, does it matter whether it works for VFs, as there are other issues with these, currently, outside of this?)

Thanks for suggesting the logic for that function! That made this an easy update, once I figured out what lines were writing the name metadata.

### DIFF
--- a/bin/gftools-add-font.py
+++ b/bin/gftools-add-font.py
@@ -179,11 +179,7 @@ def _MakeMetadata(fontdir, is_new):
 
 def _getFontFamilyName(fontfile):
   font = ttLib.TTFont(fontfile)
-  
   nametable = font['name']
-
-  help(nametable.getName)
-
   family_name = None
   typographic_name = nametable.getName(16, 3, 1, 1033)
   if typographic_name:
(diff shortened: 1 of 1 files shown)


m4rc1e:
Apologies for the delay. I'll try and look at this next with along with the fb name checks.

davelab6:
@m4rc1e what's the status of this?

m4rc1e:
Keep this open until we fix the name checks in fontbakery.

davelab6:
@m4rc1e were the name checks fixed?

m4rc1e:
Nope. Haven't had time.

davelab6:
@m4rc1e still pending?

PR #129, opened 2019-03-06
gftools-fix-hinting: added.

m4rc1e:
Enable bit 3 on a font's head table flags, if the font contains hinting


m4rc1e:
This should solve FB issue https://github.com/googlefonts/fontbakery/issues/2338

PR #132, opened 2019-04-26
Update variable font build script for public sans

eliheuer:
The Public Sans project is looking for a fontmake build script in this issue: https://github.com/uswds/public-sans/issues/84

So I am updating this to the latest version in case they want to use it.

### DIFF
--- a/bin/gftools-build-vf.py
+++ b/bin/gftools-build-vf.py
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 # coding: utf-8
-# Copyright 2018 The Font Bakery Authors.
 # Copyright 2018 The Google Font Tools Authors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


bin/gftools-build-vf.py
```diff
@@ -180,77 +233,214 @@ def get_style_list():
...
         subprocess.call(
             "fontmake \
-                      -g sources/%s.glyphs \
-                      -o variable \
-                      --output-path fonts/%s-VF.ttf \
-            > /dev/null 2>&1"
-            % (source, source),
+                      -g source/%s.glyphs \
+                      -o ttf \
+                      --keep-overlaps -i"
```
anthrotype: Why keep overlaps for static fonts?

eliheuer: Oh, thanks! That's a mistake and shouldn't be there. I'll remove it.

This script needs some work and I think most other people I have talked to think just using BASH scripts is the better way to go. The idea for this tool was to make a very user friendly VF build script for people who use Python but might not be familiar with shell scripting.

PR #137, opened 2019-05-27
fixed missing parentheses error in fix-fstype.py

bghryct:
```
      print('font saved %s.fix' % font_path)
    else:
      print('SKIPPING: %s fsType is already 0' % font_path)
```

lines 44,46 were missing parentheses in call to print


m4rc1e:
Thanks!

PR #154, opened 2019-10-07
 add-font: Support variable fonts

m4rc1e:
Running `gftools add-font` on a variable font family should now work. I haven't gone overboard with the implementation since we'll probably need to refactor this script once we start pushing vfs with multiple axes.

Since we're not able to extract the style and weight info from a VF's filename, I've had to use the name table.

I've also tidied up `FileFamilyStyleWeight` func arg. It takes a path, not just a filename.


m4rc1e:
@felipesanches could you review this?

m4rc1e:
I'm done with this PR. Unfortunately it may stay open till next year due to the festivities.

PR #160, opened 2019-11-08
Creates OFL.txt and auto-fills DESCRIPTION.en_us.html a bit

yanone:


felipesanches:
I noticed that the Travis build failed because a dependency (glyphsLib) uses the f-string syntax that is not supported on python 2.7 This should not be a problem for us since @m4rc1e has been working on a python 3 port for gftools which I hope will be merged very soon (probably next week).

felipesanches:
Other than that, I like the feature a lot. Thanks for submitting this PR, @yanone

felipesanches:
I reviewed the code changes and they look good. Let me know if you want to make any further changes or if you're already happy to have it merged.

yanone:
Are you asking me or Marc?
I'm happy.

m4rc1e:
I just ran this pr on a family where I deleted the family's OFL.txt file. I got the following traceback

```
  File "/Users/marcfoley/Type/tools/venv/bin/gftools-add-font.py", line 289, in <module>
    app.run(main)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.7/site-packages/absl/app.py", line 299, in run
    _run_main(main, args)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.7/site-packages/absl/app.py", line 250, in _run_main
    sys.exit(main(argv))
  File "/Users/marcfoley/Type/tools/venv/bin/gftools-add-font.py", line 285, in main
    _WriteTextFile(os.path.join(ofl, ofl_text))
TypeError: _WriteTextFile() missing 1 required positional argument: 'text'
```

Issue is you're passing an `os.path.join` into the `_WriteTextFile`.

yanone:
Oops. Fixed it

felipesanches:
Good catch, @m4rc1e ! Thanks :-)

bin/gftools-add-font.py
```diff
@@ -259,10 +261,28 @@ def main(argv):
   if os.path.isfile(desc):
     print('DESCRIPTION.en_us.html exists')
   else:
-    _WriteTextFile(desc, 'N/A')
+    html = '''<p>$DESCRIPTION</p>
+
+<p>To contribute, see <a href="$UPSTREAMREPO">$UPSTREAMREPO_WITHOUT_HTTP</a></p>.
```
m4rc1e: "." shoud be within `</p>`

yanone: Yup. It was meant outside the <a>. Corrected it.

felipesanches:
@m4rc1e, could you please update us all on the state of this? It would be good to either merge or close unmerged.

davelab6:
@m4rc1e I think the check is that the file exists and isn't empty, and this is an improvement over "N/A", so I'd like to merge this.

graphicore:
It feels wrong to me to generate the OFL.txt file, the file should be located in the upstream repo and we should make an exact copy. Sure, we can ask the authors to have our version of the copyright string in it. The point is, that there's an explicit, version of the OFL, with commit history/committer in the upstream so it is traceable that the license was explicitly granted by the copyright holders.

davelab6:
> It feels wrong to me to generate the OFL.txt file, the file should be located in the upstream repo and we should make an exact copy.

This is true, when there is an upstream repo. But this tool is written for a use-case where that is not true.

So, perhaps the creation of the 2 files should be non-default actions, accessed via optional args

PR #169, opened 2019-12-18
Add gftools-fix-mvar.py script

chrissimpkins:
This script removes the MVAR table from one or more variable fonts according to fontbakery check "com.google.fonts/check/unwanted_tables" guidance:

```
 >> com.google.fonts/check/unwanted_tables
    Are there unwanted tables?
    with build/ExampleVF.ttf

      Rationale:
      Some font editors store source data in their own SFNT tables, and these
      can sometimes sneak into final release files, which should only have
      OpenType spec tables.

    * FAIL: The following unwanted font tables were found:
Table: MVAR
Reason: Produces a bug in DirectWrite which causes https://bugzilla.mozilla.org/show_bug.cgi?id=1492477, https://github.com/google/fonts/issues/2085

They can be removed by using fonttools/ttx.
```

### DIFF
--- a/bin/gftools-fix-mvar.py
+++ b/bin/gftools-fix-mvar.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 The Google Font Tools Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# See AUTHORS.txt for the list of Authors and LICENSE.txt for the License.
+#
+
+import argparse
+import os
+import sys
+
+from fontTools.ttLib import TTFont
+
+
+def main():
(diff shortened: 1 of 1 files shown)


m4rc1e:
Imo it may be better to make this more general purpose. How about giving it the ability to drop any table from a set of tables which are not critical?

chrissimpkins:
> Imo it may be better to make this more general purpose.

Love it.

> any table from a set of tables which are not critical

Have we defined this?

m4rc1e:
https://github.com/googlefonts/fontbakery/blob/abd93260405f2109ef97d36d9e40710e4c36a2b8/Lib/fontbakery/profiles/universal.py#L512-L519

chrissimpkins:
```
  UNWANTED_TABLES = {
    'FFTM': 'Table contains redundant FontForge timestamp info',
    'TTFA': 'Redundant TTFAutohint table',
    'TSI0': 'Table contains data only used in VTT',
    'TSI1': 'Table contains data only used in VTT',
    'TSI2': 'Table contains data only used in VTT',
    'TSI3': 'Table contains data only used in VTT',
    'TSI5': 'Table contains data only used in VTT',
    'prop': '', # FIXME: why is this one unwanted?
      # Marc Foley found that VFs containing a MVAR table have very
      # loose vertical metrics, even if the MVAR table hasn't adjusted
      # any vertical metric values.
    'MVAR': ('Produces a bug in DirectWrite which causes'
             ' https://bugzilla.mozilla.org/show_bug.cgi?id=1492477,'
             ' https://github.com/google/fonts/issues/2085')
  }
```

chrissimpkins:
Ok, will close this and open a new PR with that script

PR #170, opened 2019-12-18
Add gftools-fix-unwanted-tables.py script

chrissimpkins:
This PR adds the `gftools-fix-unwanted-tables.py` script.  The script removes one or more "unwanted tables" from one or more fonts.  

The supported tables are defined as the following:

```python
  UNWANTED_TABLES = {
    'FFTM': 'Table contains redundant FontForge timestamp info',
    'TTFA': 'Redundant TTFAutohint table',
    'TSI0': 'Table contains data only used in VTT',
    'TSI1': 'Table contains data only used in VTT',
    'TSI2': 'Table contains data only used in VTT',
    'TSI3': 'Table contains data only used in VTT',
    'TSI5': 'Table contains data only used in VTT',
    'prop': '', # FIXME: why is this one unwanted?
      # Marc Foley found that VFs containing a MVAR table have very
      # loose vertical metrics, even if the MVAR table hasn't adjusted
      # any vertical metric values.
    'MVAR': ('Produces a bug in DirectWrite which causes'
             ' https://bugzilla.mozilla.org/show_bug.cgi?id=1492477,'
             ' https://github.com/google/fonts/issues/2085')
  }
```

File writes are in place.

By default, all tables in the unwanted list are removed from each font path.  This can be filtered with a comma separated list of table names as an argument to the `--tables` option.

Example:

```
$ gftools-fix-unwanted-tables.py --tables MVAR,TTFA [FONTPATH 1] [FONTPATH ...]
```

Includes validation of user requests against the unwanted table definition to prevent removal of other tables.

### DIFF
--- a/bin/gftools-fix-unwanted-tables.py
+++ b/bin/gftools-fix-unwanted-tables.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+
+# Copyright 2019 The Google Font Tools Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# See AUTHORS.txt for the list of Authors and LICENSE.txt for the License.
+#
+
+import argparse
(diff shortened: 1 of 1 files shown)


felipesanches:
thanks, @chrissimpkins 
Please rebase and address any issue that may surface on Travis

felipesanches:
(we have just merged the python 3 port, so you may have to adapt to python 3 style - which you seem to already be doing at least partially)

chrissimpkins:
Thanks Felipe.  I rebased on the Py3 changes in https://github.com/googlefonts/gftools/pull/170/commits/7892274bcc9b1a4d49c51a0dc144a722b1e4fc87

This is formatted as a Py3 valid script.  Please let me know if there is anything else that you need.

felipesanches:
that's great!

PR #171, opened 2020-01-08
setup.py: Set QA dependencies as extra requirement

m4rc1e:
Many users only want the fix scripts. They do not require gftools qa.

If users install gftools without the QA dependencies and tries to run `gftools qa` an error is raised which points to our readme which has the further installation instructions.

cc @arialcrime

@chrissimpkins I will update both google/fonts and Google Sans travis files.

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -23,7 +23,6 @@ import re
 import shutil
 from collections import namedtuple
 from github import Github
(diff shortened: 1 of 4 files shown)


felipesanches:
Yes! Perfect! :-D

chrissimpkins:
Thanks Marc!

PR #181, opened 2020-02-26
fixup previous fix-isfixedpitch patch

moyogo:
The previous fixup didn’t actually fix the bit in the font.


moyogo:
@m4rc1e my apologies.

PR #188, opened 2020-03-19
Add .nam file for Hebrew Extended

danielgrumer:
This adds the Hebrew Extended character set, which extends Hebrew to fully cover Biblical Hebrew and Yiddish language. It includes full cantillation marks support, wide letters, punctuation and ligatures.


danielgrumer:
@googlebot I signed it!

PR #191, opened 2020-04-09
[util] Fix main entry point tests.

graphicore:
This is to enable running the old tests build into util/google_fonts and util/filter_lists. It's not yet clear if these are still correct though.


davelab6:
@garretrieger @rsheeter please could you take a look at this, as this is a PR on the `util` codebase which has an internal version.

rsheeter:
`google_fonts.py` has a separate unit test file that we never published, not sure what the old tests built in would be. I think this is fine to merge regardless.

graphicore:
I'd like to port these to the `Lib/gftools/tests` directory, but it's good to have the current state fixed now.

PR #192, opened 2020-04-10
Update fonts_public.proto

davelab6:
Proposal to add new source section... perhaps could be part of FamilyProto tho.

Also! Slightly different topic, but our [upstream repos sheet](https://docs.google.com/spredsheets/d/1ampzD9veEdrwUMkOAJkMNkftqtv1jEygiPR0wZ6eNl8/edit#gid=0) uses repo/branch/filepath-regex, rather than repo/commit, since we are currently most interested in the latest binaries committed by the foundry rather than a specific commit. That might also be worth including here, but per ["YAGNI"](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it) principle I didn't include those fields.

### DIFF
--- a/Lib/gftools/fonts_public.proto
+++ b/Lib/gftools/fonts_public.proto
@@ -42,5 +42,4 @@ message AxisProto {
 message SourceProto {
   optional string repository_url = 1;
   optional string commit = 2;
-  optional string filepath_regex = 3;
 }
--- a/Lib/gftools/fonts_public.proto
+++ b/Lib/gftools/fonts_public.proto
@@ -19,6 +19,7 @@ message FamilyProto {
   repeated string subsets = 8;
   optional string ttf_autohint_args = 9;
   repeated AxisProto axes = 10;
+  optional SourceProto source = 11;
 };
 
 message FontProto {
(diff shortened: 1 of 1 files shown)


davelab6:
@graphicore @m4rc1e @chrissimpkins PTAL

felipesanches:
Oh! I love this! Yes!!! :-D

felipesanches:
I should also add a FontBakery check to ensure all font projects have this field with correct info.

Lib/gftools/fonts_public.proto
```diff
@@ -38,3 +38,7 @@ message AxisProto {
   optional float max_value = 4;
 };
 
+message SourceProto {
+  optional string repository = 1;
+  optional string commit = 2;
+}
```
rsheeter, line 44: You probably want `optional SourceProto source = 11;' on FamilyProto as well

Lib/gftools/fonts_public.proto
```diff
@@ -38,3 +38,7 @@ message AxisProto {
   optional float max_value = 4;
 };
 
+message SourceProto {
+  optional string repository = 1;
```
rsheeter, line 42: repository_url if the expectation is it always will be one. If not I suggest a comment explaining what we will find here.

davelab6: Using repository_url

PR #208, opened 2020-06-04
fix-vertical-metrics: add optional output path

m4rc1e:
Current implementation prepends .fix to the file suffix e.g `FontFamily-Regular.fix.ttf`. This is messy to manipulate in bash since you have to resort to slicing.

If a user does't supply an optional output, the existing implementation is used. This will ensure we don't break existing scripts.


m4rc1e:
Just realised this script can modify a collection of fonts so this won't work.

PR #209, opened 2020-06-09
new tool: gftools packager

graphicore:
Tool to transfer font families from their upstream git repository to the google/fonts repository structure, taking care of all the details.

This introduces a new file to google/fonts, `{family_dir}/upstream.yaml`. 

~There's one hard coded `upstream.yaml` in the script, that simulates how a family that already has an `upstream.yaml`in google/fonts would behave. You can try it with "Gelasio".~
 
New dependencies in `setup.py`: `pygit2`, `strictyaml`

```
# preparation
$ export GH_TOKEN=aaabbbcccYOURTOKEN
# optional, otherwise `vi`will be used as an editor, you can also export EDITOR
$ export VISUAL=gedit
```

(updated basic usage, comprehensive docs in the making)
```
# modify a git working directory or write into an empty directory
$  gftools packager Gelasio ../googleFonts/

# `-f/--force`: allow overriding existing directories and files or branches
$  gftools packager Gelasio ../googleFonts/  -f

# `-y/----no-confirm`: don't ask interactive questions
$  gftools packager Gelasio ../googleFonts/ -f -y

# `-g/--gf-git`: create a new branch in a fork/clone of google/fonts
$  gftools packager Gelasio ../googleFonts/ -g

# use a local upstream.yaml file to package to a git branch, new or override existing
$ gftools packager gelasio.upstream.yaml ../googleFonts/ -g -f

# There's a small interactive dialogue to create and edit upstream.yaml  configuration.
# Just try any family name for that.
$ gftools packager blinker../googleFonts/ -g

# `-p/--pr`:  push a branch to google/fonts and PR (you need permissions to do so)
$ gftools packager gelasio.upstream.yaml ../googleFonts/ -g -f -p

# all options
$ gftools packager -h
```
* [x] add dependencies to setup.py (if acknowledged)
* [x] actually type-check the type-hints in packager.py
* [x] make a PR with the tool
* [x] remove "Gelasio" hardcoded upstream.yaml
* [x] write documentation
* [x] there are some possible Features also FIXMES in the `packager.py` file, decide what to do
* [x] react on feedback

### DIFF
--- a/Lib/gftools/packager.py
+++ b/Lib/gftools/packager.py
@@ -51,7 +51,7 @@ else:
 
 
 # GITHUB_REPO_HTTPS_URL = 'https://github.com/{gh_repo_name_with_owner}.git'.format
-GITHUB_REPO_SSH_URL = 'git@github.com:{gh_repo_name_with_owner}.git'.format
+GITHUB_REPO_SSH_URL = 'git@github.com:{repo_name_with_owner}.git'.format
 
 GITHUB_GRAPHQL_API = 'https://api.github.com/graphql'
 GITHUB_V3_REST_API = 'https://api.github.com'
@@ -930,8 +930,10 @@ def _check_git_target(target: str) -> None:
   except Exception as e:
     raise ProgramAbortError(f'Can\'t open "{target}" as git repository. '
                             f'{e} ({type(e).__name__}).')
-
-  remote_name_or_none = _find_github_remote(repo, 'google', 'fonts', 'master')
+  repo_owner = 'google'
+  repo_name = 'fonts'
+  repo_name_with_owner = f'{repo_owner}/{repo_name}'
+  remote_name_or_none = _find_github_remote(repo, repo_owner, repo_name, 'master')
   if remote_name_or_none is None:
     # NOTE: we could ask the user if we should add the missing remote.
     # This makes especially sense if the repository is a fork of
@@ -941,10 +943,10 @@ def _check_git_target(target: str) -> None:
     # have to download a lot of new data, as well as probably create
     # confusing situations for the user when dealing with GitHub PRs etc.
     print (f'The git repository at target "{target}" has no remote for '
-      'GitHub google/fonts.\n'
+      f'GitHub {repo_name_with_owner}.\n'
       'You can add it by running:\n'
       f'$ cd {target}\n'
-      '$ git remote add googlefonts git@github.com:google/fonts.git\n'
+      f'$ git remote add googlefonts {GITHUB_REPO_SSH_URL(repo_name_with_owner=repo_name_with_owner)}.git\n'
       'For more context, run:\n'
       '$ git remote help')
 
@@ -1093,7 +1095,7 @@ def _make_pr(repo: pygit2.Repository, local_branch_name: str,
 
   push_owner, _push_repo = push_upstream.split('/')
   pr_owner, pr_repo = pr_upstream.split('/')
-  url = GITHUB_REPO_SSH_URL(gh_repo_name_with_owner=push_upstream)
+  url = GITHUB_REPO_SSH_URL(repo_name_with_owner=push_upstream)
 
   remote_branch_name = local_branch_name
   # We must only allow force pushing/general pushing to branch names that
@@ -1558,7 +1560,7 @@ def _find_github_remote(repo: pygit2.Repository, owner: str, name: str,
   # remotes it uses.
   # If more control is needed we'll have to add it.
   accepted_remote_urls = [
-     f'git@github.com:{searched_repo}.git', # ssh
+     GITHUB_REPO_SSH_URL(repo_name_with_owner=searched_repo), # ssh
      f'ssh://git@github.com/{searched_repo}', # ssh
      f'https://github.com/{searched_repo}.git', # token (auth not needed for fetch)
   ]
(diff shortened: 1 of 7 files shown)


graphicore:
@m4rc1e please take a look.

Lib/gftools/packager.py
```diff
@@ -0,0 +1,1312 @@
...
+    'repoName': repo_name,
+    'reference': reference,
+    'oflDir': f'{reference}:ofl/{family_name}',
+    'apacheDir': f'{reference}:apache/{family_name}',
+    'uflDir': f'{reference}:ufl/{family_name}'
+  }
+
+def _get_github_api_token() -> str:
+  # $ export GITHUB_API_TOKEN={the GitHub API token}
+  return os.environ['GITHUB_API_TOKEN']
```
m4rc1e, line 171: Would you mind calling this `GH_TOKEN`?

I'm already using it here, https://github.com/googlefonts/gftools/blob/master/bin/gftools-qa.py#L327

Lib/gftools/packager.py
```diff
@@ -0,0 +1,1312 @@
...
+
+def _run_gh_graphql_query(query, variables):
+  github_api_token = _get_github_api_token()
```
m4rc1e, line 174: Personally, I don't feel we need a separate function for this, we could just call os.environ directly. If you're going to add error handling/exceptions to it then FE.

Lib/gftools/packager.py
```diff
@@ -0,0 +1,1312 @@
...
+
+# Full family name, with initial upper cases and spaces
+name:
+
+# In most cases this should be based on the GitHub https repo url:
+# this https://github.com/{{owner}}/{{repo}}.git
+repository_url:
+
+# The branch name used to update google fonts. e.g.: master
+branch:
```
m4rc1e, line 300: It would be nice to have `master` as a placeholder. Just ran the tool and forgot to fill this and it crashed.

Imo, I would prefer us to only push fonts to production that are in the master branch.

m4rc1e:
Nice stuff!

It would be great if we could store the .yaml file created by the interactive mode, even if the tool fails. It is quite disheartening to populate the form and then lose everything if you forget to fill in a field.

Lib/gftools/packager.py
graphicore: No problem, will do.

Lib/gftools/packager.py
graphicore: Not sure if we should use the environment variable, ux wise, this makes room for fallbacks or alternative methods, e.g. a preferences file.

Lib/gftools/packager.py
graphicore: master as default is good for me.

> Imo, I would prefer us to only push fonts to production that are in the master branch.

I see alternative branches as way around of having to put font binaries into the repo. That way I can make a bad practices "gf" branch, lol.

Actually, more imortant: for a QA workflow, in a feature branch, before publishing, it's good to be able to use a branch. It's not only about pushing to production.

graphicore:
> It would be great if we could store the .yaml file created by the interactive mode, even if the tool fails. It is quite disheartening to populate the form and then lose everything if you forget to fill in a field.

Agree.

> It would also be great if the tool could make a PR or even a commit as you've mentioned.

Yes, it already creates a commit in a branch, making a PR is not so much more work. Not sure it it should be contained in this tool.

> I'm guessing you may have included type hints because of strictyaml? This will be our first tool in this repo which uses them. 

The reason is that types are awesome. It's a bit similar to what unit testing is trying to achieve, but much more efficient. My code is usually typed in that way that I am always aware of the types I'm expecting in my function arguments or as return values, writing it out makes this fact just obvious and also should other help to better understand he code. At least, you can leave out the typing related tests from unit testing. When including or refactoring it's easier do the right thing and when reading the code, it's easier to to follow. Since we are now only on python > 3.5 typing is available and usable, so why don't do it.

> I'm neither for or against them but having them does require everyone who works on this repo to understand them.

No, it's opt in really. I'll make sure the `packager.py` file is actually statically checked, but no consumer of that module has to be aware of it. I.e. you can include and use it as before. The python interpreter itself does not check those types on runtime, it's really just for static checking, python 3.5 only adds the syntax to the language, and the typing module for generic types.

graphicore:
Ok, here comes a second list of TODOs:

* [ ] rename `GIT_HUB_TOKEN` to`GH_TOKEN`
* [ ] `master` as default placeholder for upstream.yaml `branch` key
* [ ] store the .yaml file created by the interactive mode, especially in the case of a crash
* [ ] create PRs (it's easy from this point)

graphicore:
Look at this: https://github.com/graphicore/googleFonts/pull/51 Note: it's a test and and an example for a multi family PR, i.e. it updates two families with one PR. This is intended mainly for super families, but also other bunch updates are possible.

davelab6:
@graphicore the first set of TODO items are still unchecked, are they actually all done?

graphicore:
@davelab6 only writing some documentation is left now. The TODO items at the top are all wrapping up tasks.

BTW new dependencies: pygit2, strictyaml, mypy see: 42bbd8a

Now I'm taking care of the user docs.

davelab6:
Awesome!

graphicore:
@m4rc1e @davelab6 please review.

[docs](https://github.com/googlefonts/gftools/tree/3e71e5727c6a46b8ce5454d4a80ba0452fdf383f/docs/gftools-packager)

m4rc1e:
Thanks @graphicore. I've had a play with it and it seems like it's ready to merge.

PR #222, opened 2020-06-24
Temporary small fix on gftools-fix-weightclass.py 

RosaWagner:
I need this script to also fix weight class in OTF italic instances, so I made a raw and simple fix. Although it is probably not the cleanest way to do it. 

A nice update would be:

- [ ] work for style name build as: weight + any other style (italic, condensed…) 
- [ ] specify these values are for otf
- [ ] add accurate values for ttf and variable


m4rc1e:
Thanks for this. I've made the changes in #223 so I'll close this,.

PR #226, opened 2020-07-16
gftools push-status added.

m4rc1e:
Determine whether the families listed in to_sandbox.txt and to_production.txt in the google fonts dir have been pushed.

### DIFF
--- a/bin/gftools-push-status.py
+++ b/bin/gftools-push-status.py
@@ -1,0 +1,169 @@
+#!/usr/bin/env python3
+# Copyright 2020 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS-IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Check push status of families on Dev, Sandbox and Production servers.
+
+This script will determine whether the families listed in
+to_production.txt and to_sandbox.txt in the google/fonts dir have been
+pushed to their respective servers from a specific date.
+
+If the files are empty, you can always use git checkout to view past
+states.
+
+
+Usage:
+
+Check push status for the past month:
+gftools push-status path/to/google/fonts
+
+Check push status for specified date:
+gftools push-status path/to/google/fonts -pd 2020-07-01
+"""
+import argparse
+from argparse import RawDescriptionHelpFormatter
+import os
+from pathlib import Path
+import gftools.fonts_public_pb2 as fonts_pb2
+from google.protobuf import text_format
+from datetime import datetime, timedelta
+import json
+import requests
+
+
+# If no push date is added we can assume that the push has probably
+# happened in the past month
+ONE_MONTH_AGO = datetime.now() - timedelta(days=31)
+
+
+def get_family_metadata(url):
+    """Get family json data from a Google Fonts metadata url"""
+    # can't do requests.get("url").json() since request text starts with ")]}'"
+    info = json.loads(requests.get(url).text[5:])
+    return {i['family']: i for i in info["familyMetadataList"]}
+
+
+def read_proto(fp, schema):
+    with open(fp, "rb") as f:
+        data = text_format.Parse(f.read(), schema)
+    return data
+
+
+def families_from_file(fp):
+    """Convert to_sandbox.txt and to_production.txt files to a list of
+    family names."""
+    results = set()
+    with open(fp) as doc:
+        family_dirs = doc.read().split()
+    metadata_files = [Path(fp).parent / d / 'METADATA.pb' for d in family_dirs]
+    missing_files = [f for f in metadata_files if not f.is_file()]
+    if missing_files:
+        raise FileNotFoundError(
+            "Following METADATA.pbs files are missing:\n{}".format(
+                "\n".join(missing_files)
+            )
+        )
+    return [read_proto(f, fonts_pb2.FamilyProto()).name for f in metadata_files]
+
+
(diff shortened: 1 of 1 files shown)


bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+import gftools.fonts_public_pb2 as fonts_pb2
+from google.protobuf import text_format
+from datetime import datetime, timedelta
+import json
+import requests
+
+
+DEV_META_URL = "http://fonts-dev.sandbox.google.com/metadata/fonts"
+SANDBOX_META_URL = "http://fonts.sandbox.google.com/metadata/fonts"
+PROD_META_URL = "http://fonts.google.com/metadata/fonts"
```
rsheeter, line 47: only the host varies, no real reason to define whole thing repeatedly. use https if possible.

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+def get_family_metadata(url):
+    """Get family json data from a Google Fonts metadata url"""
+    # can't do requests.get("url").json() since request text starts with ")]}'"
+    info = json.loads(requests.get(url).text[5:])
+    return {i['family']: i for i in info["familyMetadataList"]}
+
+
+DEV_META = get_family_metadata(DEV_META_URL)
+SANDBOX_META = get_family_metadata(SANDBOX_META_URL)
+PROD_META = get_family_metadata(PROD_META_URL)
```
rsheeter, line 59: Avoid this style of global variables. Consider making a function that fetches metadata from a url and marking it as https://docs.python.org/3/library/functools.html#functools.lru_cache

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+# distributed under the License is distributed on an "AS-IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Check push status of families on Dev, Sandbox and Production servers.
+
+This script will determine whether the families listed in
+to_production.txt and to_sandbox.txt in the google/fonts dir have been
+pushed to their respective servers within an allocated date.
```
rsheeter, line 20: "within an allocated date" is an odd turn of phrase

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+# If no push date is added we can assume that the push has probably
+# happened in the past month
+ONE_MONTH_AGO = datetime.now() - timedelta(days=31)
+
+
+def families_from_file(fp):
+    """Convert to_sandbox.txt and to_production.txt files to a list of
+    family names."""
+    results = set()
+    with open(fp, "r") as doc:
```
rsheeter, line 70: don't explicitly specify default (https://docs.python.org/3/library/functions.html#open)

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+        family_dirs = doc.read().split()
+
+    gf_path = Path(fp).parent
+    for dir_ in family_dirs:
+        metadata_file = gf_path.joinpath(f"{dir_}/METADATA.pb")
```
rsheeter, line 75: Prefer path / operator; gf_path / dir_ / 'METADATA.pb' (https://docs.python.org/3/library/pathlib.html#operators)

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+        if not metadata_file.is_file():
+            raise FileNotFoundError(f"No METADATA.pb file found in {dir_}")
```
rsheeter, line 77: It's nice to lean on comprehensions.

Consider indicating all the bad paths, e.g.

```python
metadata_files = [Path(fp).parent / d / 'METADATA.pb' for d in family_dirs]
missing_files = [f for f in metadata_files if not f.is_file()]
...raise an error reporting all the missing ones, or the first N missing ones...

return [read_proto(f, fonts_pb2.FamilyProto()).name for f in metadata_files]
```

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+    results = {"pushed": set(), "not_pushed": set()}
+    if filter_families:
+        info = {k: v for k,v in info.items() if k in filter_families}
+    for family in info:
+        last_push = info[family]['lastModified']
+        last_push = datetime.strptime(last_push, "%Y-%m-%d")
+        if last_push > push_date:
+            results['pushed'].add(family)
+        if last_push < push_date:
+            results["not_pushed"].add(family)
```
rsheeter, line 98: what if last_push == push_date?

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+    return results
+
+
+def status_reporter(server_name, status):
+    """Produce a report for a server status dict"""
+    report = []
+    report.append(f"***{server_name}***")
+    pushed = list(sorted(status["pushed"]))
+    not_pushed = list(sorted(status["not_pushed"]))
```
rsheeter, line 107: If you make a function that gives the status then you can group by status then loop over the possible status values and shove things into the report rather than having to write if/else and have variables for each status

bin/gftools-push-status.py
```diff
@@ -0,0 +1,181 @@
...
+        formatter_class=RawDescriptionHelpFormatter
+    )
+    parser.add_argument("gf_path", type=gf_repo, help="path to google/fonts dir")
+    parser.add_argument(
+        "--push_date", "-pd", type=iso_8601_to_date,
+        help="Date when last push occurred"
+    )
+    args = parser.parse_args()
+
+    push_date = args.push_date or ONE_MONTH_AGO
```
rsheeter, line 161: Can't you set a default for the argument?

m4rc1e:
@rsheeter thank you for your review. I agree with all points.

I've implemented everything you've suggested apart from the lru cache. We don't need the cache because I'm only making the request to each server once. If we do need to make more requests to the same endpoints, I'll use it, or refactor the code into a class.

bin/gftools-push-status.py
m4rc1e: Ok, I'll use "from a specific date".

m4rc1e:
@rsheeter lmk if you'd like anything added/changed and once again thank you!

PR #228, opened 2020-07-27
Gftools packager

graphicore:
minor bug fixes


graphicore:
I merged it because this should be uncontroversial.

PR #236, opened 2020-08-13
Gftools packager

graphicore:
Add the flag `--no-source` to remove the `source` key from `METADATA.pb`, because the backend systems currently can't handle the source key.


m4rc1e:
Thanks!

Issue #238, opened 2020-08-19
packager: fail to submit pr


m4rc1e:
I'm trying to package and push [Vollkorn](https://github.com/FAlthausen/Vollkorn-Typeface) using the following command:

`gftools packager "Vollkorn" /my/google/fonts/path --no-source -p`

Everything works well, apart from when it attempts to make the pr, I get the following traceback

```
DONE Creating package for Vollkorn!
Start fetching yanone/master
Traceback (most recent call last):
  File "/Users/marcfoley/Type/tools/venv/bin/gftools-packager.py", line 144, in <module>
    packager.make_package(**args.__dict__)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/gftools/packager.py", line 1649, in make_package
    _packages_to_target(tmp_package_dir, family_dirs, target, is_gf_git,
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/gftools/packager.py", line 1474, in _packages_to_target
    _packagage_to_git(tmp_package_family_dir, target, family_dir,
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/gftools/packager.py", line 1382, in _packagage_to_git
    _git_make_commit(repo, add_commit, force, yes, quiet, branch,
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/gftools/packager.py", line 1285, in _git_make_commit
    _git_fetch_master(repo, remote_name)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/gftools/packager.py", line 1773, in _git_fetch_master
    stats = remote.fetch(['refs/heads/master'], callbacks=PYGit2RemoteCallbacks())
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/pygit2/remote.py", line 128, in fetch
    payload.check_error(err)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/pygit2/callbacks.py", line 93, in check_error
    check_error(error_code)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.8/site-packages/pygit2/errors.py", line 65, in check_error
    raise GitError(message)
_pygit2.GitError: unsupported URL protocol
```

It looks like it's attempting to fetch Yanone's repo `"Start fetching yanone/master"`. Since I manage this collection, I have a tonne of remotes, `git remote -v`:

```
[...]

PR #239, opened 2020-08-19
Gftools packager: several issues

### LINKED ISSUE #233: Slim down upstream.yaml files?
@graphicore we now have 2 upstream.yaml files in the central GF repo,

* https://github.com/google/fonts/blob/master/ofl/literata/upstream.yaml
* https://github.com/google/fonts/blob/master/ofl/chilanka/upstream.yaml

I'd like to propose making these more terse, and eliminate duplication with METADATA.pb files.

```yaml
# Please edit this upstream configuration for the family accordingly.
# This is a yaml formatted file.
# An "#" (number sign) denotes a comment.
# For more help see the docs at:
# https://github.com/googlefonts/gf-docs/tree/master/METADATA
```

I'm not sure we need all these comments in every file; adding it to `/README.md` or `/TRIVIA.md` would be fine, linking to 1 reference file with it all. 

```yaml
name: Literata
repository_url: https://github.com/googlefonts/literata
category: SERIF
designer: TypeTogether
```

Are these all in METADATA.pb?

```
# Choose one of: {', '.join(CATEGORIES)}
```

Bug in the template generator :)

```
files:
  # These are some examples as comments, please modify, add, delete as necessary:
  OFL.txt: OFL.txt
  DESCRIPTION.en_us.html: DESCRIPTION.en_us.html
```

Since these have a 1:1 mapping, I think it would be good to make the key implicit and therefore dropped.

graphicore:
See #238: trying to fix wrongly selected remote for PR

Fixes an issue that happened to me when testing this, unlikely in normal usage though

I don't know if `_pygit2.GitError: unsupported URL protocol` is fixed with this!

UPDATE:

also fix #233 remove redundant info from upstream.yaml, remove comments from upstream.yaml

### DIFF
--- a/Lib/gftools/packager.py
+++ b/Lib/gftools/packager.py
@@ -148,7 +148,11 @@ def _post_github(url: str, payload: typing.Dict):
   github_api_token = _get_github_api_token()
   headers = {'Authorization': f'bearer {github_api_token}'}
   response = requests.post(url, json=payload, headers=headers)
-  response.raise_for_status()
+  if response.status_code == requests.codes.unprocessable:
+    # has a helpful response.json with an 'errors' key.
+    pass
+  else:
+    response.raise_for_status()
   json = response.json()
   if 'errors' in json:
     errors = pprint.pformat(json['errors'], indent=2)
@@ -1192,9 +1196,9 @@ def _get_change_info_from_diff(repo: pygit2.Repository, root_tree: pygit2.Tree,
     #   GIT_DELTA_UNREADABLE: X
     #   default:              ' '
     if delta.status_char() == 'D':
-      # don't look at D=deleted files, tgnore something else?
-      continue
-    all_touched_files.add(delta.new_file.path)
+      all_touched_files.add(delta.old_file.path)
+    else:
+      all_touched_files.add(delta.new_file.path)
   touched_family_dirs = set()
   for filename in all_touched_files:
     for dirname in LICENSE_DIRS:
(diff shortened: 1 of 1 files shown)


m4rc1e (changes requested):
Thank you for adding the OS X fix. We need to make another one for when we fetch the remote. I'll comment on the exact line now.

Lib/gftools/packager.py
```diff
@@ -1772,7 +1841,7 @@ def _git_fetch_master(repo: pygit2.Repository, remote_name: str) -> None:
   # using just 'master' instead of 'refs/heads/master' works as well
   stats = remote.fetch(['refs/heads/master'], callbacks=PYGit2RemoteCallbacks())
```
m4rc1e, line 1842: This doesn't work on OS X. We get the same ssh issue which your last commit fixed. I just commented out this line and replace it with

```
subprocess.run(['git', 'fetch', 'upstream'],
                             cwd="/Users/marcfoley/Type/fonts",
                             check=True,
                             stdout=subprocess.PIPE) 
```

and it worked, https://github.com/google/fonts/pull/2664.

graphicore: @m4rc1e it's done in cf7bcd4. Maybe you can give it another test run.

m4rc1e: Sure thing.

m4rc1e (approved):
LGTM. lmk if you want me to merge and I'll tag a new release for pypi.

Issue #243, opened 2020-08-23
add -o --out to fix scripts


m4rc1e:
We have many fix scripts which don't allow users to specify an output path. Most fix scripts will output a fixed font with `.fix` appended to the input font path. Adding an optional `-o/--out` arg would be a nice touch.

m4rc1e:
If we do implement this, we must add this feature so it doesn't break existing projects.

alerque:
This would be an awesome thing to have a proper fix for! I've seen hundreds of (frequently dirty) shell scripts working around this irregularity.

Orthogonality related because it causes the same genre of problem while trying to automate build pipelines: https://github.com/googlefonts/fontmake/issues/693

simoncozens:
This is something we still need to do, but it's essentially the same issue as #136.

PR #262, opened 2020-10-22
Add fix module

m4rc1e:
We have a collection of fix scripts which contain useful snippets of code that should reside in their own module imo. By converting these snippets into functions, we'll get better reuse. There's also a lot of duplication so I'd like to cut down on this.

My end goal is to keep the individual fix scripts (without changing their apis) but I also just want a single `gftools fix-font` script which should do everything.

@alerque are you opposed to this idea?

### DIFF
--- a/bin/gftools-fix-font.py
+++ b/bin/gftools-fix-font.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+"""
+gftools fix-font
+
+Update a font so it conforms to the Google Fonts specification
+https://github.com/googlefonts/gf-docs/tree/master/Spec
+"""
+import argparse
+from fontTools.ttLib import TTFont
+from gftools.fix import *
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("font", help="Path to font")
+    parser.add_argument("-o", "--out", help="Output path for fixed font")
+    args = parser.parse_args()
+
+    font = TTFont(args.font)
+
+    if "DSIG" not in font:
+        add_dummy_dsig(font)
+
+    if "fpgm" in font:
+        fix_hinted_font(font)
+    else:
+        fix_unhinted_font(font)
+
+    # TODO (Marc F) fsSelection, macStyle, usWeightClass, VF instances...
+
+    if args.out:
+        font.save(args.out)
+    else:
+        font.save(font.reader.file.name + ".fix")
+
+
+if __name__ == "__main__":
+    main()
--- a/bin/gftools-fix-font.py
+++ b/bin/gftools-fix-font.py
@@ -32,15 +32,17 @@ def main():
         fix_unhinted_font(font)
 
     if "fvar" in font:
-        remove_tables(font)
+        remove_tables(font, ["MVAR"])
 
     if args.hotfix:
         log.warning("Hotfixing fonts. Please consider fixing the source files instead")
+        remove_tables(font)
         fix_nametable(font)
         fix_fs_type(font)
         fix_fs_selection(font)
         fix_mac_style(font)
         fix_weight_class(font)
+        # TODO inherit vertical metrics if font exists on Google Fonts
 
         if "fvar" in font:
             fix_fvar_instances(font)
--- a/bin/gftools-fix-font.py
+++ b/bin/gftools-fix-font.py
@@ -6,27 +6,44 @@ Update a font so it conforms to the Google Fonts specification
 https://github.com/googlefonts/gf-docs/tree/master/Spec
 """
 import argparse
+import logging
 from fontTools.ttLib import TTFont
 from gftools.fix import *
+
+
+log = logging.getLogger(__name__)
 
 
 def main():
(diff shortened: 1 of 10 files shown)


alerque:
Opposed? No this sounds like a great idea. Provided the new `gftools fix-font` has a sensible and consistent API itself (with control over the output location) I think that will be a useful thing to have.

I would still like to see the individual fix scripts API fixed such that they all can be used similarly (see e.g. #243 where they all should accept an output arg) ... but that is just an addition not a breakage of the current API.

m4rc1e:
Great. I thought I'd ask for your thoughts since you use these scripts in Fontship.

I've just added the wip `fix-font` script to this PR. Lmk if the input/output is to your liking. Also, would you be keen to review this script/pr?

alerque:
I haven't gotten a chance to take it for a test drive yet, but reviewing the code I link the direction.

I do wonder if there might be a way to pass a list of fixes (either a do list or skip list) to in case some specific fix proves problematic. I don't have any current examples of such a situation, but re-programming to use all the individual fix scrips except the one you need to skip would be a lot harder than adding a single argument to skip something. Just a thought — like I said I haven't gotten to test drive this yet.

bin/gftools-fix-font.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("font", help="Path to font")
+    parser.add_argument("-o", "--out", help="Output path for fixed font")
+    parser.add_argument("--hotfix", action="store_true", help="Hotfix fonts.")
```
alerque, line 21: I *think* I know what this is doing (applying a certain genre of fixes that *should* be fixed in the sources in addition to the genre of fixes that are rightly post-build cleanup) but the argument itself and help text are kind of misleading. My first impression of `--hotfix` would be that the fixes were being applied inline (i.e. obsoleting the `--out` argument). Is this an established name/concept elsewhere in the tool set or is there perhaps a better name for this?

m4rc1e: > I think I know what this is doing (applying a certain genre of fixes that really ought be fixed in the sources in addition to the genre of fixes that are rightly post-build cleanup and can't be corrected in the sources)

Correct.

> but the argument itself and help text are kind of misleading.

Agreed. I still haven't thought of a decent name for this. First thought was `--force`. Open to suggestions?

> My first impression of --hotfix would be that the fixes were being applied inline (i.e. obsoleting the --out argument). Is this an established name/concept elsewhere in the tool set or is there perhaps a better name for this?

I don't think we have any scripts which inline by default. Personally, I'd still like to be able to run these fixes and not overwrite the existing font, even with this arg applied.

m4rc1e: @alerque I foolishly didn't set this PR to a draft. You'll probably uncover a lot of wtf moments for the time being. I always do a cleanup once I've got a draft down. I'd be grateful for any api suggestions/help though since you're a prominent user.

alerque: Negative on `--force`, that is a very common idiom for for an action that is likely destructive.

How about `--include-source-fixes` or something similarly verbose?

(You know you can convert PR's to Draft mode after the fact right? It's in the "Reviewers" section of the sidebar.)

m4rc1e: `include-source-fixes` is much better. If we can't think of anything better, we'll roll with it. Thanks for the draft convert as well!

m4rc1e: @alerque, slept on it and I think `--include-source-fixes` will do. Thanks.

m4rc1e:
@alerque I think I'm done.

Please note, I haven't converted all of the fix functions but I believe I have a very good base that covers most of our fix scripts. I've added tests as well because the last scripts didn't have any (I had no clue whether they'd work properly for VFs etc)

I'm still going to look for edge cases and perhaps add better exceptions if needed.

bin/gftools-fix-weightclass.py
```diff
@@ -21,32 +21,17 @@
...
-        elif desired_style.usWeightClass == 200:
-            print(f"{filename}: Updating weightClass to {275}")
-            font['OS/2'].usWeightClass = 275
-            updated = True
-    if updated:
+    fix_weight_class(font)
+    if current_weight_class != font["OS/2"].usWeightClass:
         font.save(font.reader.file.name + ".fix")
     else:
         print("{}: Skipping. Current WeightClass is correct".format(filename))
```
alerque, line 37: Not necessarily relevant to this PR, but this is one of the maddeningly inconsistent bits of the old script usages: not only is the output name hard to control, sometimes they produce output and sometimes they don't. You have to run the fix, know what file to expect, check if it exists, rename it if it does, rename the original source file if not.

alerque (approved):

m4rc1e:
Thanks for the review @alerque. I'll merge this which will then unblock me for #263 and #264.

PR #263, opened 2020-10-26
gen-stat: wip

m4rc1e:
Generate a STAT table for each font in a variable font family using the GF axis registry.


m4rc1e:
I'm making decent progress. The tool can successfully add STAT tables to any family I throw at it. Here's Amstelvar's

Screenshot 2020-11-19 at 16 22 49

I want to try and release this tomorrow.

m4rc1e:
@thlinard Any objections against dropping format 2 AxisValues since these aren't necessary in the latest version of Indesign?

Personally, I dislike format 2 axis values because they don't support linked values. This means we have to add another Regular AxisValue which is format 3 to the STAT which feels hacky.

m4rc1e:
I've kept the fvar postscript names and nameID 25 as well.

Lib/gftools/stat.py
```diff
@@ -0,0 +1,350 @@
...
+
+
+def _update_fvar_nametable_records(ttfont, stat_table):
+    nametable = ttfont["name"]
+    fvar = ttfont["fvar"]
+    family_name = font_familyname(ttfont)
+    axes_with_one_axis_value = [
+        a["values"][0] for a in stat_table if len(a["values"]) == 1
+    ]
+    tokens = [v["name"] for v in axes_with_one_axis_value]
```
m4rc1e, line 255: If there are no axes which contain only a single Axis Value, it means the font doesn't have any siblings. If this occurs, we should exit the function straight away.

m4rc1e:
I'm done. I'll wait until this evening to merge just incase anyone has any comments. I'll then cut a new release.

thlinard:
> Personally, I dislike format 2 axis values because they don't support linked values. This means we have to add another Regular AxisValue which is format 3 to the STAT which feels hacky.

Hi Marc,

Sorry, I thought I commented. You're right, the absence of linked values in format 2 is a design flaw. Without repeating the debate which had taken place on https://github.com/google/fonts/issues/2391 about the interest and usefulness of format 2 for a UI more accessible to non-specialist users, it's true that this is no longer necessary for InDesign. So OK for format 1+ format 3.

> I've kept the fvar postscript names and nameID 25 as well.

Perfect then. 😃

m4rc1e:
Thanks Thomas! I'll merge this.

PR #265, opened 2020-10-30
Add GH Action GitHub + PyPI release automation on git version tag pushes

chrissimpkins:
Adds a new GitHub Action configuration that prepares and pushes a GitHub release and PyPI source dist + wheel release when a `v` prefixed version git tag is pushed.  It can live side-by-side with the Travis CI unit testing / type checking  configuration.

@m4rc1e you will need to add the following data to the repository "Secrets" settings:

- PyPI username as `PYPI_USERNAME`
- PyPI password as `PYPI_PASSWORD`

If those data are missing when you push a version tag, you'll get a GH release but the PyPI push will fail.

This Action automatically adds the following to the body of every release:

> Please see the root of the repository for the CHANGELOG.md

You'll need to change this line to something that is appropriate here.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -1,0 +1,43 @@
+on:
+  push:
+    tags:
+      - "v*" # Push events to matching `v*` version srings. e.g. v1.0, v20.15.10
+
+name: Create and Publish Release
+
+jobs:
+  build:
+    name: Create and Publish Release
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.9]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install release dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install --upgrade setuptools wheel twine
+      - name: Create GitHub release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          body: |
+            Please see the root of the repository for the CHANGELOG.md
+          draft: false
+          prerelease: false
+      - name: Build and publish to PyPI
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: |
+          python setup.py sdist bdist_wheel
+          twine upload dist/*
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -26,9 +26,6 @@ jobs:
         id: release_notes
         run: |
           CHANGELOG=$(git tag -l --format='%(contents)' ${{ github.ref }})
-          CHANGELOG="${CHANGELOG//'%'/'%25'}"
-          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
-          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
           echo "::set-output name=changelog::$CHANGELOG"
       - name: Create GitHub release
         id: create_release
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -31,15 +31,7 @@ jobs:
           # https://github.com/actions/checkout/issues/290
           git fetch --tags --force
           TAG_NAME=${GITHUB_REF/refs\/tags\//}
-          CHANGELOG=$(git tag -l --format='%(contents)' $TAG_NAME)
-          # we have to escape all new line characters and carriage
-          # returns before passing them into set-output otherwise the
-          # body content of the release will only contain the first line
-          # https://github.com/actions/create-release/issues/38#issuecomment-640072234
-          CHANGELOG="${CHANGELOG//'%'/'%25'}"
-          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
-          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
-          echo "::set-output name=changelog::$CHANGELOG"
+          echo "$(git tag -l --format='%(contents)' $TAG_NAME)" > CHANGELOG.md
 
       - name: Create GitHub release
         id: create_release
@@ -51,10 +43,7 @@ jobs:
         with:
           tag_name: ${{ github.ref }}
           release_name: ${{ github.ref }}
-          # To set the body arg, we fetch the changelog variable we created
-          # in the previous step. Implementation is based on
-          # the third example in https://github.com/actions/create-release/pull/11
-          body: ${{ steps.release_notes.outputs.changelog }}
+          body_path: CHANGELOG.md
           draft: false
           prerelease: false
 
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -22,6 +22,14 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install --upgrade setuptools wheel twine
+      - name: Get release notes
+        id: release_notes
+        run: |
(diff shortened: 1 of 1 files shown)


m4rc1e:
This is great!

I'm not such a fan of the body placeholder text. I wonder if it's possible to instead use the content from an annotated tag instead?

m4rc1e:
I see `actions/create-release@v1` provides an optional arg called `body_path` which can be used instead of `body`. We could extract the annotated tagged release text and save it as an .md file, then load it using the `body_path` command. Let me try this out.

anthrotype:
have you seen this https://github.com/actions/create-release/pull/11

m4rc1e:
Cosimo, thank you. After reading the thread, I noticed I don't need to use the `body_path` arg. Instead I've opted to get the release notes in a separate step, then just import them into the `body` arg.

@chrissimpkins I've updated the PR. Could you take a look before I merge?

m4rc1e:
@googlebot I consent.

m4rc1e:
We can't merge this until I add the PyPi credentials.

@davelab6 can I get full permissions for this repo please? I'm currently not able to add secrets to this repo.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,47 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.9]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
```
madig, line 20: You can probably get rid of `matrix` and just use `python-version: "3.x"` here?

m4rc1e:
I tested this PR on my own branch. It produces releases which have the following appearance:

Screenshot 2020-11-09 at 14 28 23

It looks fine to me. It's a whole lot better than the non-existent releases we so far have. One slight complaint is the body text isn't formatted as markdown.

---

In order to push future releases, one can do the following:

- Create a new annotated tag which has the same version number as the version number found in `setup.py` e.g `git tag -a v0.5.0'
- Vim/nano will be opened so you can write your release notes. I'd like us to simply specify what has changed in the new release. One could argue that we could just use the commit messages. However, the commit messages are perhaps too granular in this repo. I'd prefer some human intervention here.
- Push the tags `git push --tags upstream`

This should take care of the repo's tagged release and pushing the package to PyPi.

@chrissimpkins thanks for your work on this! I may adopt this for our other tools as well since the workflow is very nice.

anthrotype:
@m4rc1e cool! will GH format the tag message as markdown?

.github/workflows/publish-release.yml
chrissimpkins: yep definitely.  sorry this was copy/pasted from blocks of my ci templates.  ty!

chrissimpkins:
> One could argue that we could just use the commit messages. However, the commit messages are perhaps too granular in this repo. I'd prefer some human intervention here.

100% agree.  A maintained, human distilled changelog is a feature for users who do not closely follow/develop a project IMO.  The ends of the spectrum from "We're always making changes to make the best product for you" to "here are every single one of the commits that were made since we last released, parse it yourself and figure out what we did" tend to be less useful to the average user who just needs to know what happened between release X and Y.

.github/workflows/publish-release.yml
chrissimpkins: @m4rc1e Please let me know if you want this change and want me to make the edit.

m4rc1e:
> @m4rc1e cool! will GH format the tag message as markdown?

This non-markdown issue seems to be my fault. As a test, I've removed my heuristic to dynamically fetch the release notes from the annotated tag and instead I've just hardcoded an md list. Well well, the hardcoded list works just fine.

Screenshot 2020-11-09 at 16 41 17

I'll look into this a bit more.

chrissimpkins:
> Well well, the hardcoded list works just fine.

I imagine this means that it will support automated PR / IR linking with URL/issue numbers in the text too?  That would be nice

chrissimpkins:
Marc, are you working on this in a fork on your personal account?  I can't find https://github.com/m4rc1e/gftools

anthrotype:
@m4rc1e maybe add a comment explaining why you have to escape all the newlines and the link to the original https://github.com/actions/create-release/pull/11#issuecomment-640071918

m4rc1e:
Ok, we can now fetch the annotated tag's message and the output is now markdown.

Screenshot 2020-11-09 at 19 00 22

This nearly killed me so here are some notes:

- https://github.com/actions/create-release/pull/11 Isn't in production yet so you must use the master instead, `uses: actions/checkout@master`, not `uses: actions/checkout@v1`
- `actions/checkout` will only fetch a single commit by default. Including`fetch-depth: 0` will fetch all branches and tags. However, annotated tag messages are not included. To include them, include `git fetch --tags --force ` in a `run` cmd, https://github.com/actions/checkout/issues/290
- Getting the version tag name required some string manipulation.

@anthrotype I will. Apologies for all the mess in this PR. I've spent a decent amount of time getting it to work. I plan to go back and tidy.

anthrotype:
thanks Marc for cracking this out. I've been wanting to get a MD-formatted GH release out of a git annotated tag for long time, never got around to solve this 👍 
we'll use this approach in all the other font tooling projects (/cc @madig)

m4rc1e:
Right, I'm done with this. I've added comments for the parts which need explaining. If anyone would like to review, feel free.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,67 @@
...
+          CHANGELOG="${CHANGELOG//'%'/'%25'}"
+          CHANGELOG="${CHANGELOG//$'\n'/'%0A'}"
+          CHANGELOG="${CHANGELOG//$'\r'/'%0D'}"
+          echo "::set-output name=changelog::$CHANGELOG"
+
+      - name: Create GitHub release
+        id: create_release
+        # create-release@v1 doesn't support the body arg so we use
+        # create-release@master instead.
+        uses: actions/create-release@master
```
chrissimpkins, line 48: I only see 3 commits to master since the [`v1` release](https://github.com/actions/create-release/releases/tag/v1) commit.  `body` should be supported as of that commit.  I am using `actions/create-release@v1` with the `body` field defined using a string literal in all of my projects.  May be worth re-pinning this @ `v1` in case they decide to push a new major release at some stage.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,67 @@
...
+          # To set the body arg, we fetch the changelog variable we created
+          # in the previous step. Implementation is based on
+          # the third example in https://github.com/actions/create-release/pull/11
+          body: ${{ steps.release_notes.outputs.changelog }}
+          draft: false
+          prerelease: false
+
+      - name: Build and publish to PyPI
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
chrissimpkins, line 63: still need to add these secrets before you attempt the first release to PyPI.

chrissimpkins (commented):
LGTM.  Thanks for all of the edits here Marc.

.github/workflows/publish-release.yml
chrissimpkins: Confirmed: https://github.com/actions/create-release/blob/0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e/action.yml#L11

madig: There's also https://github.com/actions/create-release/blob/0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e/action.yml#L14 -- maybe you want to pipe the annotation text to a file instead of doing env var acrobatics?

m4rc1e: @madig agreed, it's much better. Thank you!

.github/workflows/publish-release.yml
m4rc1e: Thanks Chris. I'll talk to Dave on Thursday to get further repo permissions.

m4rc1e:
I've added my PyPI credentials to this repo so we can merge this.

Thanks everyone for your reviews and a shout out to Chris for starting this pr.

chrissimpkins:
My pleasure. Hope that this is helpful with your releases!

madig:
You didn't delete `.travis.yml` :grimacing:

davelab6:
I love to see it

PR #268, opened 2020-11-11
Switch travis checks to github actions

m4rc1e:
Since we now use Github Actions for packaging, let's drop Travis and use it for testing as well.

Based on https://github.com/googlefonts/fontmake/pull/707. I'm not going to merge this until theirs is done first.

I'd like to look into tox and perhaps spend some time writing better tests as well.


m4rc1e:
Oh I'm expecting this to fail Windows...badly!

felipesanches:
I have a question: with tox & travis we can run the tests (& build?) locally by manually invoking tox on our development computers. Will a GHActions setup allow me to replicate the server-side steps locally as well? In general I prefer to develop a feature and test it in private before submitting a PR that will trigger the public build. I'd expect we can still keep that feature, otherwise we may be getting a bit of vendor lock-in on GitHub infrastructure with this...

alerque:
@felipesanches Yes, tox can be run locally, it isn't tied to Travis or any other remote CI. GH Actions actually allows you to do far more than Travis in this regard, you can actually setup a local test runner that uses the same containers and environments they run remotely for a full test of the CI stack locally — but for what you want you don't need to do that, you just need to run the project's test suite via Tox (or whatever it is) locally.

felipesanches:
Thanks for the clarification, @alerque

chrissimpkins:
Here are the Actions docs on Py testing with tox if helpful:

https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-python#running-tests-with-tox

m4rc1e:
I'm going to merge this.

I plan to adding tox, linting and better tests in a later PR. This solves our immediate need to escape travis.

PR #270, opened 2020-11-12
Checkout submodules into publish workflow

### LINKED ISSUE #269: Glyph data excluded in latest release
I noticed when packaging 0.5.0 for Arch Linux that the package size went from ~7MB to somewhere less than 2MB. A quick diff of the file list suggests that the most likely candidate is that two XML files in `util/GlyphsInfo/` are not in the latest release?

I'm not even sure what they were used for before or if perhaps this was intentional, but the release notes don't mention it and it seems like a major change that *might* break something for someone.

alerque:
Attempt to fix #269

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -54,4 +54,4 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
           python setup.py sdist bdist_wheel
-          twine upload dist/*
+          twine upload dist/gftools*
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -14,6 +14,7 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
+          fetch-depth: 0
       - name: Set up Python
         uses: actions/setup-python@v2
         with:
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          submodules: true
+          submodules: recursive
           fetch-depth: 0
       - name: Set up Python
         uses: actions/setup-python@v2
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -11,7 +11,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
       - name: Set up Python
         uses: actions/setup-python@v2
         with:


.github/workflows/publish-release.yml
```diff
@@ -11,7 +11,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: true
```
alerque, line 17: I'm not positive about this project, but I've run into others where Python's setuptools_scm requires regular Git history, not just a shallow checkout of the HEAD commit.

Additionally I've run into some that need the tags present, in which case you also need

```suggestion
          submodules: true
  - run: |
      git fetch --prune --tags ||:
```

.github/workflows/publish-release.yml
```diff
@@ -11,7 +11,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+          fetch-depth: 0
```
alerque, line 17: I'm not positive about this project, but I've run into others where Python's setuptools_scm requires regular Git history, not just a shallow checkout of the HEAD commit.

Additionally I've run into some that need the tags present, in which case you also need

```suggestion
          fetch-depth: 0
  - run: |
      git fetch --prune --tags ||:
```

m4rc1e:
Thanks for this and apologies for the school boy error.

I will test this PR on my own fork. I'm not sure if it's going to work but we're definitely on the right track so thanks a lot.

If I delete the GlyphInfo dir, we do in fact get tests which fail if I run `test_usage.py`. We need to ensure that the tests are run before a release is built. If any tests fail, the release shouldn't be made. We need to implement #268 in order for this to work.

alerque:
Feel free to cherry pick commits, (force) push into this branch, or whatever.

m4rc1e:
@googlebot I consent.

m4rc1e:
This should be good.

@alerque To ensure the tarballs will contain both .xml files, I made a test on my master branch which uses the Github upload artifacts action to export the tarball, you can find it here, https://github.com/m4rc1e/tools/actions/runs/366333560

alerque:
Looks agreeable to me.

alerque:
@anthrotype might know better than I, but that `twine upload dist/*` instinctively seems a little off to me. Does it know to only upload tar files? The wild card is presumably to catch the version number which you don't know easily from that CLI, but `twine upload dist/gftools-*.tar.gz` or similar might be better than pushing any other random things that get built (now or in the future) up to PyPi.

anthrotype:
> twine upload dist/gftools-*.tar.gz

agreed

chrissimpkins:
> @anthrotype might know better than I, but that `twine upload dist/*` instinctively seems a little off to me. Does it know to only upload tar files? The wild card is presumably to catch the version number which you don't know easily from that CLI, but `twine upload dist/gftools-*.tar.gz` or similar might be better than pushing any other random things that get built (now or in the future) up to PyPi.

The wildcard pushes wheels as well.  

If you don't need the wheels you can remove the `bdist_wheel` argument in https://github.com/googlefonts/gftools/blob/74421242af4eefd9dbc2cd0ed6b1789d3a30f758/.github/workflows/publish-release.yml#L53

alerque:
If what's in that folder is known to all need pushing that's fine and I'll defer, @chrissimpkins does more Python publishing than I do, the wildcard just kind of had a bad smell. Yes in this case I think you want both the sdist and bdist wheels to be published.

chrissimpkins:
> If what's in that folder is known to all need pushing that's fine and I'll defer, @chrissimpkins does more Python publishing than I do, the wildcard just kind of had a bad smell. Yes in this case I think you want both the sdist and bdist wheels to be published.

The wildcard is the documented approach for twine

https://twine.readthedocs.io/en/latest/index.html#using-twine

Fine to be more explicit IMO.  You just don't *need* to unless you are going to build the wheels and don't want to push wheels.

anthrotype:
right, you want to push both sdist and wheels. if you're really that concerned, include the name of the project in the glob, that should exclude any junk that may happen to be in dist/ dir.

m4rc1e:
Alright, I've opted to include the project name when we push to twine. Thanks to all for critiquing this PR/process, I learnt a lot.

PR #272, opened 2020-11-16
remove GlyphsInfo submodule but keep the .xml files

m4rc1e:
I haven't been able to get gh actions to package a release that includes the .xml files in the `Lib/gftools/util/GlyphsInfo` submodule. What's strange is that the files are getting pulled into the repo:

Screenshot 2020-11-16 at 14 12 17

To save my sanity, I've removed the submodule and I'll just update the files whenever they are updated. [glyphsLib](https://github.com/googlefonts/glyphsLib/tree/master/Lib/glyphsLib/data) also uses this approach. 

@alerque I've spent half a day on this so I'm opting for the easy route here.


anthrotype:
if you use [setuptools_git_ls_files](https://github.com/anthrotype/setuptools_git_ls_files) setuptools plugin to have git include all files tracked by submodules in the sdist manifest.
You just need to add `"setuptools_git_ls_files"` to the `setup_requires` keyword of your setup.py script, e.g. like we do in cffsubr [here](https://github.com/adobe-type-tools/cffsubr/blob/master/setup.py#L147-L148).

m4rc1e:
Thanks Cosimo. After adding your plugin to setup.py, the package which is generated using `setup.py sdist` will include the .xml files.

```
tar -ztvf dist/gftools-0.5.0.tar.gz
...
-rw-r--r-- runner/docker 4102534 2020-11-16 15:11 gftools-0.5.0/Lib/gftools/util/GlyphsInfo/GlyphData.xml
-rw-r--r-- runner/docker 2264108 2020-11-16 15:11 gftools-0.5.0/Lib/gftools/util/GlyphsInfo/GlyphData_Ideographs.xml
```

However, the xml files are still missing from the [release zip](https://github.com/m4rc1e/tools/releases/tag/v0.5.1). I'll dig a bit more.

m4rc1e:
Oh damn. The Github release api doesn't support sub modules, https://github.community/t/create-release-that-contains-submodule/1329

anthrotype:
> the xml files are still missing from the release zip.

you should not use github release zip, only the sdist is the canonical source distribution.
auto-generated tar.gz files from GH do not include submodules.

anthrotype:
that's not a problem, pip will only use the sdist package that you upload to PyPI, the GH archives are irrelevant and should not be discouraged. Either one clones (recursively) or downloads the sdist from PyPI.

alerque:
Yes to what Cosimo said. Use the Python generated `sdist`, that's the one that *should* be released. Besides submodules there are other things the Github "release" (which is just a `git archive` generated tarball of sources) will not include, such as preparing sources for use with setuptools (as opposed to setuptools_scm or other custom modules).

m4rc1e:
Ok cool. Thanks for clarifying this. I'll close this and fix up #270 instead.

PR #276, opened 2020-11-30
 Refactor fixes

simoncozens:
This is a collection of improvements with three main goals:

* Move auxiliary functions out of font-fixing code and into appropriate utils library
* Move font-fixing actions out of scripts and into fix.py
* Encapsulate common "load a font / fix it / save it to .fix if it changed / report any messages" logic to FontFixer class.

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -3,6 +3,7 @@ Functions to fix fonts so they conform to the Google Fonts
 specification:
 https://github.com/googlefonts/gf-docs/tree/master/Spec
 """
+from fontTools.misc.fixedTools import otRound
 from fontTools.ttLib import TTFont, newTable, getTableModule
 from fontTools.ttLib.tables import ttProgram
 from fontTools.ttLib.tables._c_m_a_p import CmapSubtable
@@ -49,6 +50,7 @@ __all__ = [
     "drop_mac_cmap",
     "fix_fsselection",
     "fix_pua",
+    "fix_isFixedPitch",
     "fix_font",
     "fix_family",
 ]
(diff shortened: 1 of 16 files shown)


m4rc1e:
Thanks Simon. I'll take a look tomorrow.

Lib/gftools/fix.py
```diff
@@ -549,30 +511,225 @@ def copy_vertical_metrics(src_font, dst_font):
...
 
 
-def family_is_vf(ttFonts):
-    has_fvar = ["fvar" in ttFont for ttFont in ttFonts]
-    if any(has_fvar):
-        if all(has_fvar):
-            return True
-        raise ValueError("Families cannot contain both static and variable fonts")
-    return False
+def fix_fsselection(ttfont, filename, usetypometrics=False):
```
m4rc1e, line 529: There's already a function to fix the fsselection called `fix_fs_selection` on L209. Imo we should use this function instead. The implementation is cleaner and it works on VFs.

Lib/gftools/fix.py
```diff
@@ -158,17 +187,18 @@ def fix_weight_class(ttFont):
     Args:
         ttFont: a TTFont instance
     """
+    old_weight_class = font["OS/2"].usWeightClass
```
m4rc1e, line 190: Rename `font` to `ttFont`

Lib/gftools/fix.py
```diff
@@ -138,7 +161,11 @@ def fix_hinted_font(ttFont):
     Args:
         ttFont: a TTFont instance
     """
+    if not 'fpgm' in ttFont:
+        return False, ["Skipping. Font is not hinted."]
```
m4rc1e, line 165: Adding this condition breaks test `Lib/gftools/tests/test_fix.py::test_fix_hinted_font`.

To fix, you can either update the test or turn the unhinted test font into a mock hinted font by inserting an `fpgm` table e.g `static_font['fpgm'] = newTable("fpgm")` on L62. The latter option seems easiest to me, otherwise we may need to include another test font which has been run through ttfa or select one which is hinted.

Lib/gftools/utils.py
```diff
@@ -239,3 +240,191 @@ def mkdir(path, overwrite=True):
...
+
+    rv = []
+#    for c in unicodedata.normalize('NFKC', smart_text(string)):
+    for c in unicodedata.normalize('NFKC', string):
+        # cat = unicodedata.category(c)[0]
+        # if cat in 'LN' or c in ok:
+        rv.append(c)
+
+    new = ''.join(rv).strip()
+    result = unidecode(new)
```
m4rc1e, line 390: This module hasn't imported this function.

simoncozens: Fixed already in b7ba4!

m4rc1e (changes requested):
Overall, very nice. +1 to moving many of the functions out of the fix module and into utils.

I'm not too convinced about the `FontFixer` class atm but I do see its purpose. Since its coupling is fairly loose, I don't mind including it. I'll play with it a bit before I make up my mind.

Please address my other comments which should fix the broken tests.

Also, would you be able to review future PRs and make further contributions to this repo? I understand you're super busy and have around another 1000 repos to maintain but we need more accountability and another dev with experience to discuss design decisions etc.

simoncozens:
Thanks; these should fix tests. (I can't run tests locally because I'm getting a weird protobuf error and they require fontforge to be installed and that messes up my Python environment, so I'm flying blind.)

The point of the FontFixer class is that the scripts currently all have their own similar-but-slightly-different pieces of code running essentially the same set of operations and that's never good news. It also nudges the "fix_..." routines towards having a common interface, which also seems worthwhile.

I'm happy to cast an eye over PRs.

Lib/gftools/utils.py
```diff
@@ -21,7 +21,8 @@
 import os
 import re
 import shutil
-from unicodedata import unidecode
+import unicodedata
+import unidecode
```
m4rc1e: I think it's `from unidecode import unidecode`

m4rc1e:
Thanks Simon!

> Thanks; these should fix tests. (I can't run tests locally because I'm getting a weird protobuf error and they require fontforge to be installed and that messes up my Python environment, so I'm flying blind.)

Ouch. Fontforge hasn't been a dependency of this repo for many years. It's probably worth making a new virtualenv and then reinstalling.

Would you like me to push a release asap?

simoncozens:
No, pilot error - it works better if I do `pytest Lib` instead of just `pytest` (which looks for any python script and imports it, causing it to die on l35 of `bin/gftools-build-font2ttf.py`). And now I've upgraded protobuf it all works. No emergency release needed, just a merge please...

PR #278, opened 2020-12-01
Grand unified build theory

simoncozens:
This has two main aims:

* To generalise the various `fix-vf-meta`/`gen_stat.py`-like scripts floating around and have a data-driven STAT table solution for arbitrary-axis variable fonts, instead of everyone copying and hacking their own Python code.
* To generalise the various `build-static.sh`/`build-vf.sh`-like scripts floating around and have a single, config-file-driven project build script.

I'm still very much working on this. The gen_stat stuff isn't there at the moment, but the second part is working nicely with Glyphs sources; I've got Texturina building from a simple JSON config file instead of a multitude of custom shell scripts.


m4rc1e:
Nice!

I've already covered the stat part, take a look at https://github.com/googlefonts/gftools/blob/master/Lib/gftools/stat.py. The function `gen_stat_tables` takes an arbitrary number of fonts which belong to a single family and will generate a STAT table for each font. It does use our Axis Registry though so it may be too restricting for some users.

simoncozens:
Oh, that's fantastic news! (And partially why I thought I'd send out a draft PR, in case I'm reinventing wheels *again*.)

So, um, why doesn't `gftools-fix-vf-meta` use it...?

m4rc1e:
> So, um, why doesn't gftools-fix-vf-meta use it...?

It's legacy. I should really add a deprecation to this script, or just rewrite it and use the stat module.

m4rc1e:
There's also `fix_family` in the fix module which should take care of calling each individual fix function for a family fonts.

Personally, I'd prefer yaml over json for the config file but this isn't a big deal.

simoncozens:
`fix_font` works better for this workflow than `fix_family` but there's not a huge difference between them. YAML is a good idea, and one I have been thinking about ever since I went with JSON (without really thinking), because I have grand plans involving GitHub Actions.

simoncozens:
To be honest the config file is *largely* unnecessary. You can also do:

```
gftools builder --family-name=Castoro UFO/Castoro-Regular.ufo UFO/Castoro-Italic.ufo
gftools builder --family-name=JetBrainsMono JetBrainsMono.glyphs JetBrainsMono-Italic.glyphs
```

and it will do the right thing. It's only if you want to customize what kinds of font are produced, where they're put, etc.
```

**Comment by simoncozens** [comment]
I'm going to leave this until Friday and we can discuss more then.

**Comment by simoncozens** [comment]
Would be worth checking if we can replicate something like https://github.com/lettersoup/Sofia-Sans/blob/master/sources/sofia_stat_table.py - may need additional `config.yaml` options.

**Comment by madig** [comment]
There's also https://github.com/daltonmaag/statmake for putting STATs into VFs.

**Comment by simoncozens** [comment]
Interesting!

`gftools.stat` is nice because it deals with the common cases for you instead of requiring you to write an explicit stylespace. Perhaps the solution would be to take a `stylespaceFile` option in the config and, if present (or if `org.statmake.stylespacePath` is present), use `statmake` for that, `gftools.stat` otherwise; that would fix the Sofia issue.

I don't like the duplication of axes between the stylespace and designspace files. Feels wrong when we have that information already. But that's partially a problem with the designspace format; it doesn't take STAT information into account, and I believe it should. And also we have the problem that designspace files only allow you to describe one variable font in a family, when there may be more than one (as in the case of regular/italic).

The *real* solution is a layer above designspace which specifies relationships between multiple designspaces, and also includes style linking information. But it's a load of effort to design and writing tooling for *yet another* variable font designspace thing. I'll just use the `stylespaceFile` thing I described above, and wait until the next UFO meeting before airing my grievances.

**Comment by simoncozens** [comment]
Oh, you have all that in designspace 5.0. Very interesting.

**Comment by madig** [comment]
Indeed. If only the imaginary v5 would be an actual thing, then I wouldn't have had to come up with statmake 😬.

**Comment by simoncozens** [comment]
It's not *completely* imaginary! You know what it looks like, all you need to do is modify designspaceLib, statmake, and fontmake to support it...

**Comment by simoncozens** [comment]
(The serious point here is that an idea for a new format is far more likely to gather support and approval if there is something close to a working implementation already.)

**Comment by madig** [comment]
If only _someone_ would make a prototype, i.e. by modifying https://gist.github.com/madig/5a3e13894f75287272a06983052b4322!

**Comment by simoncozens** [comment]
Yeah, that trick normally works on me but I'm up to my eyeballs in other stuff right now. :-)

**Comment by madig** [comment]
IIRC, yes. You give it names for all fvar stops you have in your entire project and then it tailors the list to each VF you apply it to. For full control, you need `buildStatTable()`.

**Comment by simoncozens** [comment]
Right. Now you just put the STAT axes stuff into config.yml, it will add `ital` axes where necessary and then send it to `buildStatTable`:

```YAML
sources:
  - SofiaSans-VF-New.glyphs
  - SofiaSans_Italic_VF-New.glyphs
familyName: Sofia Sans
stat:
  - name: Width
    tag: wdth
    values:
    - name: UltraCondensed
      value: 50
    - name: Condensed
      value: 75
    - name: SemiCondensed
      value: 87.5
    - name: Normal
      value: 100
      flags: 2
  - name: Weight
    tag: wght
    values:
    - name: Hairline
      rangeMinValue: 1
      nominalValue: 1
      rangeMaxValue: 50
    - name: Thin
      rangeMinValue: 50
      nominalValue: 100
      rangeMaxValue: 150
    - name: ExtraLight
      rangeMinValue: 150
      nominalValue: 200
      rangeMaxValue: 250
    - name: Light
      rangeMinValue: 250
      nominalValue: 300
      rangeMaxValue: 350
    - name: Regular
      rangeMinValue: 350
      nominalValue: 400
      rangeMaxValue: 450
      flags: 2
      linkedValue: 700
    - name: Medium
      rangeMinValue: 450
      nominalValue: 500
      rangeMaxValue: 650
    - name: Bold
      rangeMinValue: 650
      nominalValue: 700
      rangeMaxValue: 750
    - name: ExtraBold
      rangeMinValue: 750
      nominalValue: 800
      rangeMaxValue: 850
    - name: Black
      rangeMinValue: 850
      nominalValue: 900
      rangeMaxValue: 950
    - name: ExtraBlack
      rangeMinValue: 950
      nominalValue: 1000
      rangeMaxValue: 1000
```

m4rc1e:
>  it will add ital axes where necessary and then send it to buildStatTable:

It would be great if it could handle other cases such as a family which consists of a normal width font and a condensed font. In order to establish this relationship, you could define a stat for each font e.g

```
sources:
  - NormalWidth.glyphs
  - Condensed.glyphs
familyName: MyFamily
stat:
  - source: NormalWidth.glyphs
    - name: Width
      tag: wdth
      values:
      - name: Normal
        value: 100
        flags: 2
    - name: Weight
      tag: wght
      values:
      - name: Regular
        rangeMinValue: 350
        nominalValue: 400
        rangeMaxValue: 450
        flags: 2
        linkedValue: 700
      - name: Bold
        rangeMinValue: 650
        nominalValue: 700
        rangeMaxValue: 750

  - source: Condensed.glyphs
    - name: Width
      tag: wdth
      values:
      - name: Condensed
        value: 75
        linkedValue: 100
    - name: Weight
      tag: wght
      values:
      - name: Regular
        rangeMinValue: 350
        nominalValue: 400
        rangeMaxValue: 450
        flags: 2
        linkedValue: 700
      - name: Bold
        rangeMinValue: 650
        nominalValue: 700
        rangeMaxValue: 750
```

This has the added benefit that each font in a family can have different axis ranges as well.

simoncozens:
Sure, will work on.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+        if self.masters:
+            return
+        for src in self.config["sources"]:
+            if src.endswith("glyphs"):
+                gsfont = glyphsLib.GSFont(src)
+                self.masters[src] = []
+                for master in gsfont.masters:
+                    self.masters[src].append(Babelfont.open(src, master=master.name))
+            elif src.endswith("designspace"):
+                # XXX
```
m4rc1e, line 125: I take it this isn't implemented yet?

m4rc1e: If it doesn't work with designspaces yet, I don't mind if the first release is for .glyphs only. However, I'd like it to raise a `NotImplementedError` if a user supplies a designspace or we alter argv/argparse so it cannot accept a designspace.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+            newname = self.rename_variable(output_files[0])
+            self.post_process(newname)
+            all_variables[source] = newname
+        if all_variables:
+            self.gen_stat(all_variables)
+
+    def run_fontmake(self, source, args):
+        if "output_dir" in args:
+            original_output_dir = args["output_dir"]
+            tmpdir = tempfile.TemporaryDirectory()
```
m4rc1e, line 185: TemporaryDirectory can be used as a context manger e.g

```
with TemporaryDirectory() as tmpdir:
    ...
```

This approach provides automatic cleanup so L205 becomes redundant.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+            self.gen_stat_stylespace(self.config["stylespaceFile"], filenames)
+        elif "stat" in self.config:
+            self.gen_stat_config(self.config["stat"], varfonts)
+        else:
+            ttFonts = [TTFont(x) for x in filenames]
+            gen_stat_tables(ttFonts, self.config["axisOrder"])
+            for filename, ttFont in zip(filenames, ttFonts):
+                ttFont.save(filename)
+
+    def gen_stat_config(self, stat, varfonts):
```
m4rc1e, line 235: This is great. Imo, building a STAT table from a yaml file is so good it warrants its own function in `gftools.stat`. I can then reuse it in `gftools gen-stat` by adding an optional file argument.

I can happily do this at a later point.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+            else:
+                if "italic" in f.lower():
+                    stat[-1] = ital_stat_for_italic
+                else:
+                    stat[-1] = ital_stat_for_roman
+                this_stat = stat
+            buildStatTable(ttFont, this_stat)
+            ttFont.save(f)
+
+    def gen_stat_stylespace(self, stylespaceFile, filenames):
```
m4rc1e, line 268: Perhaps we should log a warning that we prefer the other methods over this one, since they provide more functionality?

m4rc1e (commented):
Overall, this is great. All comments I've raised are just niceties.

Just a few things to add before we can merge:
- Add the missing dependencies to both setup.py and requirements.txt
- Since this tool is a wrapper, we don't need extensive unit tests for each method/func but I would like to at least see an end to end test provided in `Lib/gftools/tests/test_usage.py`.

I'd also like us to dog food it before merging. It would be great to have @rosawagner @vv-monsalve and myself use it on a few families before hitting the merge button.

Thanks Simon! it is a pleasure working with you and reading your code.

Lib/gftools/builder/__init__.py
simoncozens: It should work with designspaces. This code is just auto-guessing the familyName if the user doesn't supply it. I'll make that clearer.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,341 @@
...
+    def __init__(self, configfile=None, config=None):
+        if configfile:
+            self.config = yaml.load(open(configfile), Loader=yaml.SafeLoader)
+            if os.path.dirname(configfile):
+                os.chdir(os.path.dirname(configfile))
+        else:
+            self.config = config
+        self.masters = {}
+        self.logger = logging.getLogger("GFBuilder")
+        self.fill_config_defaults()
```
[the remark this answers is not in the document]
simoncozens: Yeah, I'm not sure that would be worthwhile; for one thing, the `fill_config_defaults` still needs to happen since some config parameters are computed from other ones, which you can't just do with keyword args. Moving yaml out does make some sense. But meh. It's fine.

Lib/gftools/builder/__init__.py
simoncozens: I looked at that, but we only need a temporary directory when we're producing a directory full of stuff (as opposed to, say, an individual variable font). There's no problem making a temp directory we're not going use, but is it really worth it to save a line? :-)

Lib/gftools/builder/__init__.py
simoncozens: I think let's do that now.

m4rc1e (commented):
You also need to include fontmake, babelfont, ttfautohint-py and pyyaml. Add these to the setup.py as well.

Lib/gftools/builder/__init__.py
```diff
@@ -0,0 +1,312 @@
...
+            self.gen_stat(all_variables)
+
+    def run_fontmake(self, source, args):
+        if "output_dir" in args:
+            original_output_dir = args["output_dir"]
+            tmpdir = tempfile.TemporaryDirectory()
+            args["output_dir"] = tmpdir.name
+
+        if source.endswith(".glyphs"):
+            FontProject().run_from_glyphs(source, **args)
```
anthrotype, line 190: I wouldn't recommend to use fontmake as a library. FontProject's constructor modifies logging global state (calls logging.basicConfig which will set a handler on the root logger the first time it gets called, then subsequently does nothing). I keep saying I want to change the API but never get to actually do it, I know..
Anyway, I think of fontmake as primarily a command line tool. It's "api" are its command line options, and the kind output it leaves behind on the filesystem. The "library" that does (or should do) the bulk of the business is ufo2ft (but I know there's still logic that's only available in fontmake itself, e.g. the fontmake.instantiator to interpolate static UFOs).
So I "reserve the right" to change the python interface of fontmake.font_project module without making any commitments its stability. In light of that, it would be advisable for your application as client of fontmake to treat it as a console tool and call it as a subprocess. But feel free to ignore me.

anthrotype: FYI https://github.com/googlefonts/fontmake/issues/719

simoncozens: OK, I hear what you're saying about the lack of API guarantees from `FontProject`. But I think it's worth the "risk". (Note also that Pilcrow takes the same approach of calling `FontProject` directly.)

There's a lot of logic in fontmake itself. Setting up the correct ufo2ft calls, connecting with glyphsLib, interpolating static UFOs... all of this is important stuff, and that's why it takes FontProject 1200 lines of code to do it. It's by no means a thin wrapper around ufo2ft.

I suppose we could retool to call `fontmake.__main__.main`. Perhaps we will do so when we need to. API guarantees are never *really* contracts - a new version of *anything* could change the API, and we would have to work around it. For now it works. If there are changes down the line, we deal with them.

simoncozens: Yeah, doing that already.

m4rc1e:
@simoncozens shall I take another look and merge if good? 

Since you're hammering glyphs3 support, I can get this across the line if you want?

simoncozens:
Thanks; it should be fine, but please review and do with it what you will. I am up to my eyeballs in Glyphs3 and it really isn't my idea of fun, so it would be great if you could take this off my plate!

m4rc1e:
@googlebot I consent.

m4rc1e:
Thanks Simon.

I've just added a few commits which add a functional test, formatting and importing a few more missing dependencies.

It seems we may have an issue regarding output paths (or me just nitpicking). If I have the following config

```YAML
sources:
  - Lora.glyphs
outputDir: "."
familyName: Lora
```

The output will be saved relative to my cwd. Wouldn't it be better if the output was relative to the config file path instead?

simoncozens:
Good point!

simoncozens:
It does already chdir to the configfile though in l.89. Is that not working somehow?

m4rc1e:
It didn't work for me. I discovered this when I added the test case.

m4rc1e:
Strange but I'm no longer able to replicate this, I'll merge and if it resurfaces, I'll fix it.

felipesanches:
Thanks, guys! This new feature will likely make upstream font repo builds much simpler. I'm excited about it and I'll soon start trying to use it on some font repos.

PR #279, opened 2020-12-06
fix typo

hyvyys:
I'm using this as an excuse to ask why this is even needed? Should the default figures be tabular (or maybe even tabular old-style)? (Hardly reasonable).


hyvyys:
@googlebot I signed it!

PR #280, opened 2020-12-08
WIP gftools.html aka diffbrowsers v2

m4rc1e:
Currently, if we want to test fonts on different browsers, we have to use [diffbrowsers](https://github.com/googlefonts/diffbrowsers) and [gfregression](https://github.com/googlefonts/gfregression). I want to deprecate both of these tools and replace them with the two new tools being developed in this pr, `gftools gen-html proof` and `gftools gen-html diff`.

---

But why?

- GFR's first release was in 2016. At the time, Browserstack couldn't do local testing so we needed a server we could push fonts to. Nowadays, this isn't an issue as Browserstack's local testing is pretty decent. 

- Needing a server increased the dependencies significantly. GFR uses Google Cloud Platform, Flask, Docker, docker-compose and a document database. This requires developers who also have devop skills.

- Servers cost $. At a measly $50 a month, I know this is peanuts but if we want performance which is the same as our local systems, this costs a lot more.

- Only I can push updates since I'm the only one who can access to the Google Cloud Platform account

---

Both tools in this PR will simply generate the html test documents on the user's machine. If they want to test the pages on different browsers, a local daemon server is started and then browserstack is run in local mode.

I've spent a fair amount of time on the design (still very much a wip) so hopefully other foundries/developers can roll their own templates or template generator classes.

cc @madig. I remember you looking at GFRegression years ago so this may be of interest to you folks.


m4rc1e:
@simoncozens would you mind taking a look?

I'm just about done. I'll do another pass where I clean up variable names/add doc strings etc. Would also be interested in your thoughts on the overall design.

Lib/gftools/html.py
```diff
@@ -0,0 +1,542 @@
...
+    1: "50%",
+    2: "62.5%",
+    3: "75%",
+    4: "87.5%",
+    5: "100%",
+    6: "112.5%",
+    7: "125%",
+    8: "150%",
+    9: "200%",
+}
```
m4rc1e, line 58: Move to gftools.util.styles

simoncozens:
Looks very good to me. The only tiny niggle I have is l.88. We've got, like, a hundred different ways of telling if a font is regular or italic, in different places, and this adds another one. Otherwise I like it.

m4rc1e:
> The only tiny niggle I have is l.88. We've got, like, a hundred different ways of telling if a font is regular or italic, in different places, and this adds another one.

Ok cool, I'll fix this.

PR #282, opened 2020-12-21
Use ttfs filenames instead of src filenames when building stat tables from config

m4rc1e:
I've refactored the `stat.gen_stat_tables_from_config` function so we don't need to supply source files. Instead, we just specify the ttf filenames e.g 

Current:

```YAML
Family.glyphs:
- name: Weight
  tag: wght
...
```

This pr
```YAML
Family[wght].ttf
- name: Weight
  tag: wght
...
```

By doing this, I can reuse the function in `gftools gen-stat` without users needing to add the sources (in some projects, we may not even have sources).

cc @simoncozens. Still wip, will add some tests. I've also tidied up some functions since many of them were saving files in place so they didn't have good reuse etc.


simoncozens (approved):
Nice - much cleaner way to do it!

PR #284, opened 2020-12-24
Refactor gftools push-status

m4rc1e:
I decided to refactor this script since we also need to check if the push files contain typos. I could've forced this feature in but the code was already a mess.

If you supply the arg `--lint`, it will check if all paths specified in the server files are valid directories e.g 
`gftools push-status path/to/google/fonts --lint`. I will hook this command up to the google/fonts ci as well.

If our push pipeline increases in complexity, we should probably move these functions to their own dedicated module etc. For now, I think it's still fine as a script.


davelab6:
Nice work Marc

PR #296, opened 2021-01-21
Builder: Add includeSourceFixes to config file

m4rc1e:
The builder will post process built fonts by calling the `gftools.fix.fix_font` function. This function has an additional arg called `include_source_fixes` which will hotfix a font (we advise that users fixes the sources). This PR will allow users to specify in their config file that they want to enable these fixes.


m4rc1e:
I also fixed an issue in `gftools.fix.fix_fvar_instances`. Since a dict is mutable, it was assigning the same coordinates to each instance which is clearly wrong.

Issue #290, opened 2021-01-14
[builder] Clean up UFO with option to keep it


RosaWagner:
We trying the builder with Marc and Vivi and noticed the builder doesn't clean up masters/ufo and instances/ufo by itself.
reminder to @m4rc1e : add the ufo clean up feature by default, with an option to keep these directories.

RosaWagner:
And when the building is finished, a small end note like "Complete" or "All Done!" would really be *the cherry on the cake*.

RosaWagner:
@moyogo I see the end message, but the master_ufo and instance_ufo are still there. There is an empty instance_ufo directory, but there is also a non-empty one in master_ufo.

![Capture d’écran 2021-01-27 à 18 42 41](https://user-images.githubusercontent.com/12222436/106031446-92a95600-60cf-11eb-8c4a-31b68fa1eb63.png)

PR #299, opened 2021-01-22
Gftools builder cleanup

moyogo:
Fixes https://github.com/googlefonts/gftools/issues/290

- [x] Add log message to say "All done!"
- [x] Add --no-clean-up argument and clean up `instance_ufos/`  by default

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -53,6 +53,7 @@ required, all others have sensible defaults:
 * ``ttDir``: Where to put TrueType static fonts. Defaults to ``$outputDir/ttf``.
 * ``otDir``: Where to put CFF static fonts. Defaults to ``$outputDir/otf``.
 * ``woffDir``: Where to put WOFF2 static fonts. Defaults to ``$outputDir/webfonts``.
+* ``cleanUp`: Whether or not to remove temporary files. Defaults to ``true``.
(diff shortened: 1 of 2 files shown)


Lib/gftools/builder/__init__.py
```diff
@@ -249,11 +261,13 @@ def gen_stat_stylespace(self, stylespaceFile, varfonts):
                 stylespace, ttFont, additional_locations
             )
 
-    def build_static(self):
+    def build_static(self, cleanup=True):
```
m4rc1e, line 264: Do we need to run the cleanup step if we generate a VF as well?

m4rc1e: There's no need to create the `cleanup` function argument. Just use `self.config['cleanup']` in the function body instead.

Lib/gftools/builder/__init__.py
```diff
@@ -254,11 +267,13 @@ def gen_stat_stylespace(self, stylespaceFile, varfonts):
...
         self.build_a_static_format("otf", self.config["otDir"], self.post_process)
         if self.config["buildWebfont"]:
             self.mkdir(self.config["woffDir"], clean=True)
         self.build_a_static_format("ttf", self.config["ttDir"], self.post_process_ttf)
+        if cleanup:
```
m4rc1e, line 275: Use `if self.config['cleanup']` instead.

m4rc1e:
Thanks!

PR #304, opened 2021-01-23
add-font: get actual font name for metadata rather than splitting file name

felipesanches:
This is a followup to pull request #126 that was automatically closed by mistake when this project renamed its **"master"** branch to **"main"**.

**Be sure to also read all the comments posted in the original PR (#126)**

Original PR description as submitted on Feb 25, 2019 by @thundernixon:
----

This aims to resolve https://github.com/googlefonts/gftools/issues/122

I've added Marc's suggested code to get the font nameID 16, else nameID 1, and this results in a more-accurate result for font names in metadata generated by `gftools-add-font.py`.

I have a few questions before I'll feel confident in this merge:

- **Is there a cleaner way to add this functionality?** E.g. should the new `_getFontFamilyName` function be placed somewhere else, such as in `util/google_fonts.py`?

- **Will this work for VFs?** Might it accidentally include a style name, because nameID 1 will likely include the default style name? (Also, in the context of this update, does it matter whether it works for VFs, as there are other issues with these, currently, outside of this?)

Thanks for suggesting the logic for that function! That made this an easy update, once I figured out what lines were writing the name metadata.

----
**Be sure to also read all the comments posted in the original PR (#126)**

### DIFF
--- a/bin/gftools-add-font.py
+++ b/bin/gftools-add-font.py
@@ -179,11 +179,7 @@ def _MakeMetadata(fontdir, is_new):
 
 def _getFontFamilyName(fontfile):
   font = ttLib.TTFont(fontfile)
-  
   nametable = font['name']
-
-  help(nametable.getName)
-
   family_name = None
   typographic_name = nametable.getName(16, 3, 1, 1033)
   if typographic_name:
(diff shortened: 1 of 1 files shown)


felipesanches:
@googlebot I consent.

thlinard:
>     * **Will this work for VFs?** Might it accidentally include a style name, because nameID 1 will likely include the default style name? (Also, in the context of this update, does it matter whether it works for VFs, as there are other issues with these, currently, outside of this?)

If, by retrieving a family name by name ID 16, else nameID 1, one get a name which isn't stripped of all reference to a style, the font is broken and should not be released, VF or not.

PR #309, opened 2021-01-23
wip: gen-statics added

felipesanches:
This is a followup to pull request #264 that was automatically closed by mistake when this project renamed its **"master"** branch to **"main"**.

**Be sure to also read all the comments posted in the original PR (#264)**

Original PR description as submitted on Oct 26, 2020 by @m4rc1e:
----

Generate static fonts from a variable font using its STAT table.

I can't merge this until #262 is finished. I'll then need to cleanup and remove some of the duplication in this script.

----
**Be sure to also read all the comments posted in the original PR (#264)**


m4rc1e:
First update in a while.

My new plan is to allow designers to include static instances in the builder config files e.g:

```YAML
sources:
  - MavenPro.glyphs
includeSourceFixes: true
outputDir: "../fonts"
instances:
  MavenPro[wght].ttf:
  - styleName: "Regular"
    coordinates:
      wght: 400
  - styleName: "Medium"
    coordinates:
      wght: 500
  - styleName: "SemiBold"
    coordinates:
      wght: 600
  - styleName: "Bold"
    coordinates:
      wght: 700

```

cc @vv-monsalve, @RosaWagner

m4rc1e:
@simoncozens I think this is ready, mind taking a quick look? 

This feature is useful because it allows users to generate static fonts which are independent from the designspace source files. At Google Fonts, we only allow 18 fvar instances which only include wght + ital axes e.g "Bold", "Bold Italic". In fonts like Recursive which have a tonne of cool axes, having only 18 fonts isn't great. One could argue that we could have separate designspace sources for the static fonts but I feel this is inferior and doesn't remove the duplication.

moyogo:
@googlebot I consent.

m4rc1e:
@googlebot I consent

Lib/gftools/instancer.py
```diff
@@ -0,0 +1,42 @@
+from fontTools.varLib.instancer import instantiateVariableFont, OverlapMode
+from gftools.fix import update_nametable, fix_fs_selection, fix_mac_style
+
+
+__all__ = ["gen_static_font"]
+
+
+def gen_static_font(
+    var_font, family_name, style_name, axes, keep_overlaps=False, dst=None
+):
```
moyogo, line 10: The signature could be `gen_static_font(var_font, axes, family_name=None, style_name=None, keep_overlaps=Fase, dst=None)` considering instantiateVariableFont() will have a `family_name` parameter in https://github.com/fonttools/fonttools/pull/2189.

Should the `axes` argument be called `axis_values`?

Lib/gftools/builder/__init__.py
```diff
@@ -42,9 +55,12 @@
 * ``logLevel``: Debugging log level. Defaults to ``INFO``.
 * ``stylespaceFile``: A statmake ``.stylespace`` file.
 * ``stat``: A STAT table configuration. This may be either a list of axes and
-    values as demonstrated above, or a dictionary mapping each source file to a
+    values as demonstrated above, or a dictionary mapping each variable font to a
     per-source list. If neither ``stylespaceFile`` or ``stat`` are provided, a
     STAT table is generated automatically using ``gftools.stat``.
+* ``instances``: A list of static font ttf instances to generate from each variable
+    font as demonstrated above. If this argument isn't provided, static ttfs will
+    be generated for each instance that is specified in the source files.
```
moyogo, line 63: Either uppercase TTF or avoid mentioning TTF  as we do for buildVariable for example.

Lib/gftools/builder/__init__.py
```diff
@@ -306,6 +323,36 @@ def build_static(self):
...
         if self.config["buildTTF"]:
             self.build_a_static_format("ttf", self.config["ttDir"], self.post_process_ttf)
+            if "instances" in self.config:
+                self.instantiate_static_fonts(
+                    self.config["ttDir"], self.post_process_ttf
+                )
+            else:
+                self.build_a_static_format(
+                    "ttf", self.config["ttDir"], self.post_process_ttf
+                )
```
moyogo, line 333: The first `self.build_a_static_format()` can be removed. I think it comes from a rebase/merge conflict.

moyogo (changes requested):

Lib/gftools/builder/__init__.py
m4rc1e: Good catch. Thanks.

Lib/gftools/instancer.py
m4rc1e: The fonttools PR will update the name table but there's a chance it won't comply with our specification. At GF, if a static family features axes which aren't wght or ital, it will be released as a [new family](https://fonts.google.com/?query=condensed). The fonttools pr doesn't do this. It will simply update nameIDs 2 and 17 and leave the family names intact.

I've just checked our [update_nametable](https://github.com/googlefonts/gftools/blob/main/Lib/gftools/fix.py#L375) and it's capable of updating the family name if non wght/ital particles exist. This means I can update the signature as you're suggesting.

m4rc1e: Thanks.

I've updated the signature as per your recommendation. However, we still need to update the name table using our own name table updater since GF style names can only have wght+ital tokens. The fontTools instancer doesn't have this restriction.

> Should the axes argument be called axis_values?

This may cause some confusion since some users may think they need to supply STAT AxisValues, which isn't what we're doing. Perhaps `axis_coordinates` is a better arg name?

Tests are failing because we've just merged the fontTools PR. Once a new release is cut, the tests should pass.

moyogo:
LGTM

PR #312, opened 2021-01-23
Builder: Add includeSourceFixes to config file

felipesanches:
This is a followup to pull request #296 that was automatically closed by mistake when this project renamed its **"master"** branch to **"main"**.

**Be sure to also read all the comments posted in the original PR (#296)**

Original PR description as submitted on  by @m4rc1e:
----

The builder will post process built fonts by calling the `gftools.fix.fix_font` function. This function has an additional arg called `include_source_fixes` which will hotfix a font (we advise that users fixes the sources). This PR will allow users to specify in their config file that they want to enable these fixes.

----
**Be sure to also read all the comments posted in the original PR (#296)**


m4rc1e:
Thanks for reopening this!

PR #315, opened 2021-01-25
builder: Fix getting familyName if config does not specify one

### LINKED ISSUE #314: Builder: crashes if no family name is provided
Just disccovered this whilst generating Maven Pro, with the following config:

```
sources:
  - MavenPro.glyphs
includeSourceFixes: true
```

```
  File "/Users/marcfoley/Type/tools/bin/gftools-builder.py", line 56, in <module>
    builder = GFBuilder(configfile=args.file[0])
  File "/Users/marcfoley/Type/tools/Lib/gftools/builder/__init__.py", line 94, in __init__
    self.fill_config_defaults()
  File "/Users/marcfoley/Type/tools/Lib/gftools/builder/__init__.py", line 149, in fill_config_defaults
    familynames = set([x.info.familyName for x in self.masters.values()])
  File "/Users/marcfoley/Type/tools/Lib/gftools/builder/__init__.py", line 149, in <listcomp>
    familynames = set([x.info.familyName for x in self.masters.values()])
AttributeError: 'list' object has no attribute 'info'
```

If I include the familyName myself it is fine.

I'll fix this then cut a new release.

m4rc1e:
Currently, `set([x.info.familyName for x in self.masters.values()])` doesn't work because `self.masters.values()` is a nested list structured as `[src1->[master1, master2], src2->[master1, master2]]`. To fix, I do another loop in `self.masters.values()` so i access the individual masters. 

Fixes #314 

cc @simoncozens.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -146,7 +146,12 @@ class GFBuilder:
                 self.config["familyName"] = designspace.sources[0].familyName
             else:
                 self.load_masters()
-                familynames = set([x.info.familyName for x in self.masters.values()])
+
+                familynames = set()
+                for masters in self.masters.values():
+                    for master in masters:
+                        familynames.add(master.info.familyName)
+
                 if len(familynames) != 1:
                     raise ValueError(
                         "Inconsistent family names in sources (%s). Set familyName in config instead"


Lib/gftools/builder/__init__.py
```diff
@@ -146,7 +146,12 @@ def fill_config_defaults(self):
                 self.config["familyName"] = designspace.sources[0].familyName
             else:
                 self.load_masters()
-                familynames = set([x.info.familyName for x in self.masters.values()])
+
+                familynames = set()
```
m4rc1e, line 150: More than happy to make this into a list comp but I've found list comps containing 2 loops are difficult to debug.

PR #319, opened 2021-01-27
builder: Fix cleanUp

m4rc1e:
`instance_ufo` and `master_ufo` should be deleted when `cleanUp` is set to true. Fixes https://github.com/googlefonts/gftools/issues/290#issuecomment-768457334

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -278,7 +278,8 @@ class GFBuilder:
             self.mkdir(self.config["woffDir"], clean=True)
         self.build_a_static_format("ttf", self.config["ttDir"], self.post_process_ttf)
         if self.config["cleanUp"]:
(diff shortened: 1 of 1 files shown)


m4rc1e:
I'm having some doubts about the current implementation.

Currently, we will end up deleting source files if the project is ufo based and they are stored in the dirs `master_ufo` etc. I think we need to check what type the sources are first, then run the clean up operations which are suitable for the file type. I'll add this now.

m4rc1e:
@moyogo I've changed the implementation so it will cleanUp regardless of what source format is used. Can you review this please? (I'm not able to select you as a reviewer)

moyogo:
@m4rc1e LGTM I'm sorry I didn't get those folders earlier but now I do. Good catch on the already existing folders.

m4rc1e:
Thanks Denis!

PR #327, opened 2021-02-12
Add flatten components argument. Fixes #325

### LINKED ISSUE #325: [builder] Flatten components
Please include flattening of components by default into the builder.

My local code adjustment worked fine:
```
    def run_fontmake(self, source, args):
        args["flatten_components"] = True
```

But I wasn't sure if that's the right place, so I refrained from making a PR.
Also, maybe consider to make it optional, defaulting to `True`.

simoncozens:


m4rc1e:
Thanks LGTM. Tests are failing due to a previous commit which i will fix.

simoncozens:
I'm already on it!

Issue #297, opened 2021-01-22
[builder] Roman/Italic variation issue in Indesign when stat in config.yaml


RosaWagner:
The italic is going back to roman when using the weight slider in Indesign — but only when there is the stat in the config.yaml.

https://user-images.githubusercontent.com/12222436/105488609-2f27be80-5cb2-11eb-945b-be6a06aa88ca.mov

These are my different test:
- When exported as we use too ie. with **gen-stat.py + build.sh** : *no issue* 
See: https://github.com/RosaWagner/Texturina/tree/master/fonts/variable
- When exported with **builder + stat entry in config.yaml** : *issue*
See: https://github.com/RosaWagner/Texturina/tree/master/fonts-builder/variable
- When standard font exported with **builder without stat entry in config.yaml**: **no issue**
See: https://github.com/RosaWagner/Archivo/tree/master/fonts/variable

RosaWagner:
This is the config.yaml, maybe I am missing something:

```
    sources:
      - Texturina.glyphs
      - Texturina-Italic.glyphs
    axisOrder:
      - opsz
      - wght
    outputDir: "../fonts-builder"
    familyName: Texturina
    buildStatic: false
    stat:
      Texturina[opsz,wght].ttf:
      - name: Optical size
        tag: opsz
        values:
        - name: 12pt
          value: 12
          flags: 2
        - name: 72pt
          value: 72
      - name: Weight
        tag: wght
        values:
        - name: Thin
          value: 100
        - name: ExtraLight
          value: 200
        - name: Light
          value: 300
        - name: Regular
          value: 400
          linkedValue: 700
          flags: 2
        - name: Medium
          value: 500
        - name: SemiBold
          value: 600
        - name: Bold
          value: 700
        - name: ExtraBold
          value: 800
        - name: Black
          value: 900
      - name: Italic
        tag: ital
        values:
        - name: Roman
          value: 0
          linkedValue: 1
          flags: 2
      Texturina-Italic[opsz,wght].ttf:
      - name: Optical size
        tag: opsz
        values:
        - name: 12pt
          value: 12
          flags: 2
        - name: 72pt
          value: 72
      - name: Weight
        tag: wght
        values:
        - name: Thin
          value: 100
        - name: ExtraLight
          value: 200
        - name: Light
          value: 300
        - name: Regular
          value: 400
          linkedValue: 700
          flags: 2
        - name: Medium
          value: 500
        - name: SemiBold
          value: 600
        - name: Bold
[...]

PR #329, opened 2021-02-16
Update fvar instance ps names when using gen_stat_tables_from_config

m4rc1e:
@RosaWagner Please check if this PR solves #297


RosaWagner:
It works ! Thanks !!

m4rc1e:
I'll cut a release later

PR #330, opened 2021-02-22
[builder] Add buildOTF config parameter

### LINKED ISSUE #326: [builder] Make static OTFs optional
OTFs are worthless in GF workflow and generating them eats away precious life time. 
Please consider to diversify the options to generate OTFs and TTFs separately.

moyogo:
This fixes #326 ([\[builder\] Make static OTFs optional](https://github.com/googlefonts/gftools/issues/326)) by adding a `buildOTF` parameter, and also adds a `buildTTF` parameter


m4rc1e (changes requested):
Personally, I think both otfs and ttfs should be built by default. This is something we've always done. I do understand that it takes longer but the upstream repo should have a few more niceties than just what we need. Other than that, LGTM.

m4rc1e:
Thanks. LGTM

PR #334, opened 2021-03-01
builder: use familyName to construct VF filename

m4rc1e:
Currently, if we generate a variable font, the filename will inherit the source filename e.g `Myfont_roman.designspace --> Myfont_roman[wdth,wght].ttf`. What we really want is the generated font's family name with no spaces e.g `Roboto Serif --> RobotoSerif[wdth,wght].ttf`.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -90,7 +90,7 @@ from fontmake.font_project import FontProject
 from ufo2ft import CFFOptimization
 from gftools.fix import fix_font
 from gftools.stat import gen_stat_tables, gen_stat_tables_from_config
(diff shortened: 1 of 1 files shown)


m4rc1e:
Hmm, trying to automagically work out the filename seems like a good idea but there are too many edgecases one needs to consider. Atm, if a family consists of two VFs, one for Roman, the other Italic, one of the fonts will get overwritten. One could account for these but where do you stop. What happens if there's a separate set of sources for Condensed etc.

I think I'll close this PR but keep some of the fixes for the instances.

m4rc1e:
Closing this and opening a more aptly named pr for the commits I'd like to keep.

PR #341, opened 2021-03-08
Fix vf filename round 2

m4rc1e:
Currently, we use the axisOrder parameter to construct VF filenames. This doesn't work correctly since a VF filename must only list the axes which are in the font, not all the axes which make up a family.

This issue becomes apparent when you have a VF family which consists of multiple fonts e.g:

Current implementation:
`MyFamily[wdth,wght,ital].ttf`
`MyFamily-Italic[wdth,wght,ital].ttf`

Neither font has an `ital` axis in the fvar tables so they shouldn't be included in the filename 

This pr will produce the following:
`MyFamily[wdth,wght].ttf`
`MyFamily-Italic[wdth,wght].ttf`

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -283,11 +283,11 @@ class GFBuilder:
             return [os.path.join(original_output_dir, x) for x in file_names]
 
     def rename_variable(self, fontfile):
-        if "axisOrder" not in self.config:
-            font = TTFont(fontfile)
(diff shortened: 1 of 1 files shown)


simoncozens (approved):

m4rc1e:
Thanks Simon!

RosaWagner:
@m4rc1e does this mean that the "ital" axis should be specified in the axisOrder ?

m4rc1e:
>does this mean that the "ital" axis should be specified in the axisOrder ?

Yes, if the family has it.

The axisOrder parameter should specify all the axes which exist in the family, not just a single font.

PR #351, opened 2021-03-23
builder: axisOrder defaults to fvar order

moyogo:
config["axisOrder"] may be undefined, this makes it default to the fvar order as mentioned in the doc.


m4rc1e:
Thanks!

Issue #345, opened 2021-03-16
Builder: publish YAML schema files for config


m4rc1e:
gftools builder shouldn't run if build/config include parameters which are misspelled or not valid. Bonus points for suggesting fixes to spelling mistakes e.g `buildotf is not a valid parameter. Did you mean buildOTF?`

m4rc1e:
@eliheuer don't know if you'd like to work on this? we can pair on it if you'd like? This isn't urgent but it's definitely something we'll need. Perhaps we'll tackle it once you start using the builder.

eliheuer:
That would be great! I have a meeting Friday I'm busy getting ready for but I'll work on this after that.

eliheuer:
@m4rcle I added a draft PR for this here: https://github.com/googlefonts/gftools/pull/356

I spent some time reviewing Python error handling over the weekend, so pairing on this would be helpful if you have time. Thanks!

jenskutilek:
Maybe related: Would it be possible to supply YAML schema files for the config yaml file? Something like https://github.com/googlefonts/gftools/blob/main/Lib/gftools/builder/schema.py but in yaml format itself. That would simplify editing config.yaml in editors which use the schema, e.g. VSCode.

This could also be used to validate any config.yaml when/before running the builder.

eliheuer:
@jenskutilek Maybe that would be something the Google Fonts Unified Repository could include? https://github.com/googlefonts/Unified-Font-Repository

jenskutilek:
It could even reside in a central location, VSCode can take a URL for yaml schemas used for validation:

https://gist.github.com/jenskutilek/42415ea048896a4cde81e5e5f9147b7b#file-settings-json-L58-L60

```json
{
  "yaml.schemas": {
    "https://gist.githubusercontent.com/jenskutilek/1f93723fd3aed89810d61d5d6934b518/raw/6bd5f08614ad3368c8f3a66e86dd91dfd107380d/fontfabrik-v1.yaml": "*.fofa.yaml"
  }
}
```

simoncozens:
We've kind of done this using strictyaml, although now under builder2 build config files can be a lot more flexible.

PR #356, opened 2021-04-06
Builder: Add error handling to check config files

eliheuer:
As suggested in issue #345. I need some feedback to make sure I'm going in the right direction.

So far, if I change `sources` to something like `sssources` in `sources/config.yaml`, I get the following error that includes the original traceback and some more helpful error messages giving hints on how to fix the problem:
```
$ gftools builder sources/config.yaml

  File "/Users/Air/GF-2021/python/gftools/bin/gftools-builder.py", line 86, in <module>
    builder.build()
  File "/Users/Air/GF-2021/python/gftools/Lib/gftools/builder/__init__.py", line 145, in build
    self.build_variable()
  File "/Users/Air/GF-2021/python/gftools/Lib/gftools/builder/__init__.py", line 238, in build_variable
KeyError: 'sources'

Error: A key in the configuration file, typically ``config.yaml``, is likely misspelled.
Error caused by key: 'sources'
Possibly misspelled key in the configuration file:  ['sssources']
```

This is not currently working with some of the other keys, I'm looking into that now...

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -84,39 +84,61 @@ required, all others have sensible defaults:
 
 """
 
-from fontTools import designspaceLib
-from fontTools.ttLib import TTFont
+from babelfont import Babelfont
 from fontmake.font_project import FontProject
-from ufo2ft import CFFOptimization
+from fontTools import designspaceLib
+from fontTools.otlLib.builder import buildStatTable
+from fontTools.ttLib import TTFont
+from fontTools.ttLib.woff2 import main as woff2_main
+from gftools.builder.schema import schema
 from gftools.fix import fix_font
 from gftools.stat import gen_stat_tables, gen_stat_tables_from_config
 from gftools.utils import font_is_italic, font_familyname, font_stylename
 from gftools.instancer import gen_static_font
-from fontTools.otlLib.builder import buildStatTable
+from strictyaml import load, YAMLError
+from ufo2ft import CFFOptimization
+import difflib
+import glyphsLib
+import logging
+import os
+import re
+import shutil
 import statmake.classes
 import statmake.lib
-from babelfont import Babelfont
 import sys
-import os
-import shutil
-import glyphsLib
 import tempfile
-from fontTools.ttLib.woff2 import main as woff2_main
-import logging
-import yaml
 
 
 class GFBuilder:
     def __init__(self, configfile=None, config=None):
-        if configfile:
-            self.config = yaml.load(open(configfile), Loader=yaml.SafeLoader)
-            if os.path.dirname(configfile):
-                os.chdir(os.path.dirname(configfile))
-        else:
-            self.config = config
-        self.masters = {}
-        self.logger = logging.getLogger("GFBuilder")
-        self.fill_config_defaults()
+        self.load_builder_config(configfile, schema)
+
+    def load_builder_config(self, configfile, schema=schema):
+        with open(configfile) as f:
+            unprocessed_yaml = f.read()
+        try:
+            if configfile:
+                self.config = load(unprocessed_yaml, schema).data
+                if os.path.dirname(configfile):
+                    os.chdir(os.path.dirname(configfile))
+            else:
+                self.config = config
+            self.masters = {}
+            self.logger = logging.getLogger("GFBuilder")
+            self.fill_config_defaults()
+        except YAMLError as error:
+            error_problem = str(error.problem)
+            bad_key = str(re.findall(r"'(.*?)'", error_problem))
+            config_keys = re.findall(r'"(.*?)"', str(schema._validator))
+            config_keys.extend(re.findall(r"'(.*?)'", str(schema._validator)))
+            key_close_matches = difflib.get_close_matches(bad_key, config_keys, 8, 0.4)
+            raise YAMLError(
+                f"\nERROR **********************"
+                f"\nA key in the configuration file, typically ``config.yaml``, is likely misspelled."
+                f"\nError caused by key: {bad_key}"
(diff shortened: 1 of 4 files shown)


m4rc1e:
Thanks for the PR and good start!

I really like it can detect typos. However, I think it would be better if we used something like [strictyaml](https://github.com/crdoconnor/strictyaml) as we do for the packager since it allows us to validate user config files against a schema. Totally happy to pair on this at some point if you'd like?

eliheuer:
Sure thanks! I'll look into the strictyaml stuff in the packager and update this PR. I can pair on this whenever works for you.

eliheuer:
@googlebot I consent.

eliheuer:
@m4rc1e I pushed an update to this that is ready for review. 

I will push another update to include a test but wanted to post this in case you have time to give feedback before I write the test.

eliheuer:
This is what the output looks like now. I mostly tested this on Texturina, but will try it with a few more when I write the test.
```
$ gftools builder sources/config.yaml

while parsing a mapping
  in "<unicode string>", line 10, column 1:
    sstat:
    ^ (line: 10)
unexpected key not in schema 'sstat'
  in "<unicode string>", line 84, column 1:
          value: '1'
    ^ (line: 84)

ERROR ****************************************************************

A key in the configuration file, typically ``config.yaml``, is likely misspelled.

Error caused by key: ['sstat']
Possibly misspelled key in the configuration file: ['stat']
```

Lib/gftools/builder/schema.py
```diff
@@ -0,0 +1,55 @@
+from strictyaml import load, Map, MapPattern, Str, Int, Seq, YAMLError, Optional, Bool
```
m4rc1e, line 1: Wrap these in parenthesis 

```
from strictyaml (
    load,
    Map,
    ...
)
```

Include a one sentence doc string as well.

m4rc1e: or just run black on the file.

bin/gftools-builder.py
```diff
@@ -72,12 +80,33 @@
...
 if args.dump_config:
-    import sys
     import yaml
 
     with open(args.dump_config, "w") as fp:
         config= {k: v for (k, v) in builder.config.items() if v is not None}
         fp.write(yaml.dump(config, Dumper=yaml.SafeDumper))
     sys.exit()
 
+try:
```
m4rc1e, line 90: We're checking the yaml after the builder has been initialised (see L64) when we actually want to check it during initialisation. What we want to do is replace current yaml loading routine in `Lib/gftools/builder/__init__.py L112` with your code below. I'd also wrap this code into its own helper function e.g

```PYTHON
def load_builder_config(fp, schema=schema):
    ....
```

bin/gftools-builder.py
```diff
@@ -72,12 +80,33 @@
...
+    config_yaml = load(unprocessed_yaml, schema)
+except YAMLError as error:
+    print(error)
```
m4rc1e, line 93: I like you've retained the spell checking. This may be too verbose for power users/devs but for font designers etc it is perfect.

Instead of calling print, we should just call `YAMLError` just once e.g:

```
except YAMLError as error:
    raise YAMLError(
        f"'\nERROR ' + '*'*64 + '\n'
        f"A key in the configuration file, typically ``config.yaml``, is likely misspelled."
        f"\nError caused by key: {bad_key}"
        ...
    )

**Review by m4rc1e — CHANGES_REQUESTED** [review]
See my changes above. Can pair as well.

File: Lib/gftools/builder/schema.py (hunk 1 above)
Comment (eliheuer) [reply Lib/gftools/builder/schema.py]: Thanks. This is fixed in commit 9463326 that I just pushed to this PR. Black `v20.8b1` doesn't format it this way for me, for some reason.

File: bin/gftools-builder.py (hunk 2 above)
Comment (eliheuer) [reply bin/gftools-builder.py]: Thanks. This is done in commit 9463326.

File: bin/gftools-builder.py (hunk 3 above)
Comment (eliheuer) [reply bin/gftools-builder.py]: Thanks. This is done in commit 9463326.

**Comment by eliheuer** [comment]
@m4rc1e The requested changes have been made and this is ready for review again at commit 9463326.

I also added one additional commit (649425f) adding a test, but I'm not very confident this is the best way to do this. Feel free to discard this commit if everything else looks good to merge. I'm also happy to pair on this at any time.

Thanks!

**Comment by m4rc1e** [comment]
@googlebot I consent.

File: Lib/gftools/builder/__init__.py [hunk 4]
```diff
@@ -118,6 +122,29 @@ def __init__(self, configfile=None, config=None):
...
+    def load_config(self, configfile):
+        with open(configfile) as f:
+            unprocessed_yaml = f.read()
+        try:
+            return load(unprocessed_yaml, schema).data
+        except YAMLValidationError as e:
+            if "unexpected key not in schema" in e.problem:
+                bad_key = str(re.findall(r"'(.*?)'", e.problem))
+                config_keys = re.findall(r'"(.*?)"', str(schema._validator))
+                config_keys.extend(re.findall(r"'(.*?)'", str(schema._validator)))
```
m4rc1e, line 134: Our next task is to remove the regular expressions.

eliheuer: This is done in commit 298895b

eliheuer:
@m4rc1e I removed the regular expressions as requested. I couldn't figure out how to do this while retaining all the original functionality of the spell checker, but better done than perfect.

This is ready for review again, thank you!

m4rc1e:
Cool. I will test and merge tomorrow. Thank you for working on this!

m4rc1e (approved):
Fabulous thank you!

Let's improve this a bit more in the future!

PR #357, opened 2021-04-08
Packager: Fix typo in packager -B help

eliheuer:
```
Changes "Allow executing the bash command in stored in the "build" key of upstream-conf, if present."
to      "Allow executing the bash command stored in the "build" key of upstream-conf, if present."
```


m4rc1e:
Thanks!

PR #362, opened 2021-04-22
Sync nam files with internal copy

rsheeter:


davelab6:
@graphicore @m4rc1e @moyogo ptal

m4rc1e (approved):
LGTM.

Nit: some nam file mention they they've been taken from https://github.com/google/fonts/blob/master/tools/encodings when this dir no longer exists. We should probably just remove these comments.

PR #368, opened 2021-06-02
builder: use DecomposeTransformedComponents filter

### LINKED ISSUE #367: Builder: use DecomposeTransformedComponents filter
Transformed components are problematic in variable fonts and with hinting.
Gftools builder should probably use the ufo2ft DecomposeTransformedComponents filter by default.

moyogo:
Fixes #367.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -81,6 +81,7 @@ required, all others have sensible defaults:
 * ``axisOrder``: STAT table axis order. Defaults to fvar order.
 * ``familyName``: Family name for variable fonts. Defaults to family name of first source file.
 * ``flattenComponents``: Whether to flatten components on export. Defaults to ``true``.
+* ``decomposeTransformedComponents``: Whether to decompose transformed components on export. Defaults to ``true``.
 
 """
 
@@ -98,6 +99,7 @@ from gftools.instancer import gen_static_font
 from strictyaml import load, YAMLError
 from strictyaml.exceptions import YAMLValidationError
 from ufo2ft import CFFOptimization
+from ufo2ft.filters import loadFilterFromString
 import difflib
 import glyphsLib
 import logging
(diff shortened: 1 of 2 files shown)


Lib/gftools/builder/__init__.py
```diff
@@ -281,7 +285,22 @@ def run_fontmake(self, source, args):
...
+
+        if (
+            self.config["flattenComponents"] or
+            self.config["decomposeTransformedComponents"]
+        ):
+            filters = args.get("filters", [])
+            if self.config["flattenComponents"]:
+                filters.append([
+                    loadFilterFromString("FlattenComponentsFilter")
+                ])
```
m4rc1e, line 297:
```suggestion
            if self.config["flattenComponents"]:
                filters.append(
                    loadFilterFromString("FlattenComponentsFilter")
                )
```

You're appending a list which contains a single filter. What you want is to just append the filter.

Lib/gftools/builder/__init__.py
```diff
@@ -281,7 +285,22 @@ def run_fontmake(self, source, args):
...
+
+            if self.config["decomposeTransformedComponents"]:
+                filters.append([
+                    loadFilterFromString("DecomposeTransformedComponentsFilter")
+                ])
```
m4rc1e, line 302:
```suggestion
                filters.append(
                    loadFilterFromString("DecomposeTransformedComponentsFilter")
                )
```

Same here :-)

m4rc1e: Currently, ufo2ft doesn't like the lists being appended. If I remove them, then it's fine.

m4rc1e: Here's a minimal test case

[component_test.zip](https://github.com/googlefonts/gftools/files/6584956/component_test.zip)

moyogo: Thanks! It's fixed.

m4rc1e (approved):
LGTM. Thanks!

m4rc1e:
I'll fix the CI in another pr.

PR #375, opened 2021-06-14
make designer dirs ascii

### LINKED ISSUE #371: add-designer: make sure created directories are ascii only
Accent are added to designer folders.

m4rc1e:
Fixes #371

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -36,6 +36,7 @@ if sys.version_info[0] == 3:
     from configparser import ConfigParser
 else:
     from ConfigParser import ConfigParser
+import unicodedata as uni
 
 # =====================================
 # HELPER FUNCTIONS
@@ -504,3 +505,7 @@ def read_proto(fp, schema):
         data = text_format.Parse(f.read(), schema)
     return data
 
+
+def strip_accents(string):
+    """Remove accents from a string e.g 'àce' --> 'ace'"""
+    return "".join([uni.normalize("NFD", x)[0] for x in string])
--- a/bin/gftools-add-designer.py
+++ b/bin/gftools-add-designer.py
@@ -34,6 +34,7 @@ from PIL import Image
 from gftools.designers_pb2 import DesignerInfoProto
 from google.protobuf import text_format
 from pandas.core.base import PandasObject
+from gftools.utils import strip_accents
 
 
 def process_image(fp):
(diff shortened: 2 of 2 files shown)


m4rc1e:
This will still fail for certain combos such as 

```
>>> strip_accents("Ǽplas").isascii()
False
```

simoncozens:
Is this designer directory exposed to the public anywhere? How is it used? I feel like doing something fragile like this will end up being a time bomb which will go off at some later date when a font gets added that's designed by 杉山誠 or युनिवर्सल थर्स्ट. I'd feel happier just using an MD5sum or some other kind of hash.

m4rc1e:
> Is this designer directory exposed to the public anywhere?

Designer dirs are here https://github.com/google/fonts/tree/main/catalog/designers

> How is it used?

Each directory contains the metadata for each font author.

Thinking about it, I'm not sure why we need to ensure that these dirs are ascii only. AFAIK win, mac and linux all support non-ascii names. I guess we want it because the current dirs are all ascii.

m4rc1e:
@simoncozens After emailing the eng team, it seems like we're stuck with ascii folder names and we can only have Latin names. Due to these requirements, I'm going to do the following:

- Raise an exception is a user enters a name which isn't Latinized
- Drop accents and also convert æ to ae etc for dirs

felipesanches:
we were asked not to use unidecode due to its license. Please see https://github.com/googlefonts/fontbakery/issues/3316 and https://github.com/googlefonts/fontbakery/issues/3306

felipesanches:
on that PR I included a very crude implementation of the bare minimal needed for ascii-fying the designer names:
https://github.com/googlefonts/fontbakery/pull/3318/files

It would be good if we used the same code snippet on both tools

m4rc1e:
Thanks Felipe! I'll drop the dependency.

Since we need this in both projects, perhaps we should roll our own with an Apache license?

cc @davelab6

simoncozens:
The original Unidecode tables by Sean Burke are available under the Perl Artistic License. But perhaps a suitable workaround would be to use something Felipe's bare minimum, and then raise an exception if the designer's name still isn't ASCII after bare-minimum processing, because it indicates that a name has got into the directory that shouldn't be in there.

(I still think you should just use a hash. But anyway.)

moyogo:
There's https://pypi.org/project/text-unidecode/ under Artistic License (Perl) 1.0 as another option.

m4rc1e:
I've made a package which is basically just a large hashmap which covers the basic multilingual plane using the Sean Burke data. I'll bring it up on Friday.

PR #376, opened 2021-06-16
builder: add support for STAT format4 axisValues

### LINKED ISSUE #333: builder: support format 4 STAT tables
It seems we cannot specify format 4 stat tables for the builder.

cc @vv-monsalve

m4rc1e:
Fixes #333

Users will need to declare format 4 axisValues in a separate section in their builder files. This mimics the implementation in fontTools where format4 axisValues are declared in a different variable in `buildStat`.

```
stat:
  - name: Weight
    tag: wght
    values:
    - name: Regular
      value: 500
    - name: Bold
      value: 700
statFormat4: # <--- new section
  - name: "Echo Fox"
    location:
      wght: 400
  - name: "Echo Space"
    location:
      wght: 700
```

A stat table also needs to be declared in order for this to work.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -30,6 +30,15 @@ which looks like this::
           nominalValue: 1
           rangeMaxValue: 50
         ...
+    statFormat4:
+      - name: Green
+        location:
+          wght: 300
+          wdth: 200
+      - name: Blue
+        location:
+          wght: 400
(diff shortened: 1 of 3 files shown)


moyogo:
It would be nicer if we had been able to keep the data under `stat` but the structure is too different, otherwise it looks good to me and works as expected.

moyogo (approved):

PR #378, opened 2021-06-16
Smooth all ppem sizes

aaronbell:
The current GASP setting in the fix_unhinted_font tooling actually turns on hinting across the entire range. An unhinted font should have gridfit turned off, and smoothing turned on. This fix corrects that.

Additionally, once the hinting has been turned off, the prep table modification has no impact on font rendering, so that code has been removed as well.

### DIFF
--- a/Lib/gftools/builder/_ninja.py
+++ b/Lib/gftools/builder/_ninja.py
@@ -63,19 +63,20 @@ class NinjaBuilder(GFBuilder):
                 os.remove(temporary)
 
         # Clean up temp build files
-        search_directory = os.getcwd()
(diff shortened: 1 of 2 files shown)


m4rc1e:
Thanks for the pr. Unfortunately, I think the current implementation is correct. I've implemented something similar in fonttools-rs and copied the current implementation here due to the following comment/thread, https://github.com/googlefonts/fontbakery/issues/2672#issuecomment-722027792

aaronbell:
Hm. The PR aligns with the recommendation I've always received from Microsoft for how to treat unhinted fonts. Perhaps it is worth doing some tests.

m4rc1e:
+1 to tests. Please also look at how the fonts perform on a browser which uses GDI since our current implementation is optimised for this.

Issue #380, opened 2021-06-29
builder: append gftools version number to font's unique font id name record


m4rc1e:
This will be handy for debugging purposes. It also means we can write an FB check to test whether users are generating fonts with the latest version of gftools.

@eliheuer would you be interested in writing this?

eliheuer:
Sure, I'll give this a try. Thanks!

eliheuer:
Ok, it looks like all I need to do is update the `unique_name` function in `Lib/gftools/utils.py`, PR on the way.

PR #383, opened 2021-07-01
Builder: append gftools version number to font's unique font id name record, fixes issue #380

eliheuer:
Not yet working, but I'm starting with a draft PR if anyone wants to give feedback. Should fix issue #380 when done.

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -26,7 +26,7 @@ from unidecode import unidecode
 import browserstack_screenshots
 from collections import namedtuple
 from github import Github
-from pkg_resources import resource_filename
+from pkg_resources import resource_filename, get_distribution
 from google.protobuf import text_format
 import time
 import json
@@ -325,7 +325,8 @@ def unique_name(ttFont, nameids):
     font_version = _font_version(ttFont)
     vendor = ttFont["OS/2"].achVendID.strip()
     ps_name = nameids[6]
-    return f"{font_version};{vendor};{ps_name}"
+    gftools_version = get_distribution('gftools').version
+    return f"{font_version};{vendor};{ps_name};{gftools_version}"
 
 
 def _font_version(font, platEncLang=(3, 1, 0x409)):


anthrotype:
I don't think the TrueType Unique ID is the right place for these sort of things. That one is fairly standard and always includes the font name, its version and the vendor. These three price of info are sufficient to uniquely identify fonts. Extra pieces of info like the tooling versions don't belong there. If a different version of gftools made a difference, then the font version would have been bumped as well, which would be reflected in the TT unique ID.

Makeotf traditionally appended its own version at the end of name id 5 (the version string).

Some vendors also use the nameID 10 (description) for this sort of things.

In a similar fontmake issue, people have proposed to use the meta table which allows to define private tags that can contain any string.

m4rc1e:
> If a different version of gftools made a difference, then the font version would have been bumped as well, which would be reflected in the TT unique ID.

This may not be the case. Users often don't have the latest version of gftools installed or they may hand the work over to another designer who is using a newer or older version.

I agree that the uniqueID may not be the correct record for this info though.

m4rc1e:
If we're not going to make a meta table with our own private tag. Appending the info to the version record (nameID 5) is better imo.

eliheuer:
Thank you both! I will try to update this soon, looking at the open type [spec](https://docs.microsoft.com/en-us/typography/opentype/spec/name) it seems like either will work.

Name ID 5:
>Version string. Should begin with the syntax “Version <number>.<number>” (upper case, lower case, or mixed, with a space between “Version” and the number)... A character such as “;” is helpful to separate different pieces of version information.

Name ID 10:
>Description; description of the typeface. Can contain revision information, usage recommendations, history, features, etc.

At this point gftools builder is not putting anything in Name ID 10, and I get this in Name ID 5:
``` xml
<namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
    Version 1.003; ttfautohint (v1.8.3)
</namerecord>
```
So I'll try to update this PR to get this, unless anyone thinks Name ID 10 would be better:
``` xml
<namerecord nameID="5" platformID="3" platEncID="1" langID="0x409">
    Version 1.003; ttfautohint (v1.8.3); gftools (v0.7.3)
</namerecord>
```

m4rc1e:
Closing since we have a better implementation in #780

PR #389, opened 2021-07-08
Update .gitignore to ignore PyCharm and VSCode caches

twardoch:


requirements.txt
```diff
@@ -1,4 +1,4 @@
-fonttools[ufo]
+fonttools[ufo,lxml,woff,unicode,interpolatable]
```
m4rc1e, line 1: Could you move the dependency updates to a new pr please?

May be worth discussing these first.

PR #392, opened 2021-07-12
Ensure fsSelection bit 7 is set

simoncozens:
fontbakery is [now](https://github.com/googlefonts/fontbakery/issues/3241) demanding the OS/2.fsSelection bit 7 be set. When fixing the fsSelection field, we should set it.


m4rc1e (changes requested):
For CJK fonts, we don't want bit 7 enabled. FB will skip them, https://github.com/chrissimpkins/fontbakery/blob/5552759e9710d6c5c8b9a8d48b9c94e6ebfc6fe8/Lib/fontbakery/profiles/googlefonts.py#L5318

PR #395, opened 2021-08-02
Modularize gftools add-font

m4rc1e:
@rsheeter I need to use the metadata generation functionality in Fontbakery. In order to this I've moved it into its own module. Do you have any objections?


rsheeter:
No objection. gftools has diverged so far from the internal version that they feel like they aren't related anymore.

PR #397, opened 2021-08-04
Clean up dependencies

m4rc1e:
Add extra install requirements for building fonts and building designer profiles.

cc @felipesanches. I'll let you know when I'm done what dependencies will be added to Fontbakery when we eventually import this repo.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,7 +24,7 @@ jobs:
           python-version: ${{ matrix.python-version }}
       - name: Install packages
         run: |
-          pip install .
+          pip install .[all]
           pip install mypy==0.812 pytest
       - name: Run Tests
         run: |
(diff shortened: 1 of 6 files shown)


m4rc1e:
@alerque sorry for not letting you know sooner but we're going to use gftools as a dependency in fontbakery, https://github.com/googlefonts/fontbakery/pull/3404#issuecomment-891936430. This will be a breaking change if projects haven't pinned to a particular version of gftools. If you need to update your repos to have the same functionality, you can do `pip install gftools[font-build, designer-build]`, to install all dependencies you can run `pip install gftools[all]`.

m4rc1e:
Putting this on hold since we may want a bigger and better sledgehammer.

alerque:
Thanks for the heads up.

And yes, getting this split up into optional extra features so that using it for font building vs. using it for designer info pages may have different requirements sounds like the right way to handle it. Without that you'd foist a bunch of cruft into places it doesn't need to be.

m4rc1e:
Closing since this pr is stale and we're still thinking about how best to tackle this in the future.

PR #402, opened 2021-08-19
Add "meta" table generator

simoncozens:
See googlefonts/fontbakery#3418

- Add langcodes dependency to handle BCP47 codes
- Code to create meta tables with language tag information
- Plug meta table generator into builder


Lib/gftools/meta.py
```diff
@@ -0,0 +1,75 @@
...
+from langcodes.tag_parser import LanguageTagError
+from langcodes.data_dicts import VALIDITY
+
+__all__ = ["gen_meta_table"]
+
+
+log = logging.getLogger(__name__)
+
+
+def gen_meta_table(ttFont, config):
```
m4rc1e, line 21: It may be nice to move this up a level to the fontTools builder

m4rc1e: I have no problems with keeping it here if there's little interest.

simoncozens: This way allows for the possibility of adding a command-line utility to add the table, outside of the builder.

Lib/gftools/meta.py
```diff
@@ -0,0 +1,75 @@
...
+    try:
+        code = Language.get(value)
+    except LanguageTagError:
+        log.warning(f"Skipping unparsable script/lang tag '{value}'")
+        return False
+    if not code.script:
+        log.warning(
+            f"meta table script/lang tags must have a script component, '{value}' does not"
+        )
+        return False
```
dscorbett, line 62: `Language.get('Latn')` returns a tag with the language `latn` and no script, making this function incorrectly return false.

Lib/gftools/meta.py
```diff
@@ -0,0 +1,75 @@
...
+    if code.is_valid():
+        return True
```
dscorbett, line 64: `is_valid` returns true for some invalid codes like `aaj-Latn` and `en-Latnx`.

Lib/gftools/builder/__init__.py
```diff
@@ -71,6 +77,9 @@
     values as demonstrated above, or a dictionary mapping each variable font to a
     per-source list. If neither ``stylespaceFile`` or ``stat`` are provided, a
     STAT table is generated automatically using ``gftools.stat``.
+* ``meta``: A meta table configuration. This should be a dictionary with ``dlng``
+    and/or ``slng`` keys, with the values being a list of script/language tags
+    as defined in the OpenType Specification.
```
dscorbett, line 82: This might be misleading because they aren’t script tags and language system tags in the usual OpenType sense.

Lib/gftools/meta.py
simoncozens: Do you have a positive suggestion?

simoncozens:
Let’s keep this out of the next release until the tag issues are fixed. Who’da thought that OpenType having its own bespoke almost-but-not-quite-BCP47 language tag system would have caused such problems?

Lib/gftools/meta.py
dscorbett:
```python3
if len(code.language) == 4 and code.language.isalpha():
    code.script = code.language.title()
    code.language = None
```

Lib/gftools/meta.py
```diff
@@ -0,0 +1,76 @@
...
+        log.warning(f"Invalid language code '{code.language}' in '{value}'")
+    if not VALIDITY.fullmatch(code.script):
+        log.warning(f"Invalid script code '{code.script}' in '{value}'")
+    if code.territory and not VALIDITY.fullmatch(code.territory):
+        log.warning(f"Invalid territory code '{code.territory}' in '{value}'")
+    if code.variants:
+        for variant in code.variants:
+            if not VALIDITY.fullmatch(variant):
+                log.warning(f"Invalid variant code '{variant}' in '{value}'")
+    return code.is_valid()
```
dscorbett, line 76: This returns true for the invalid codes `zh-zho-Hans` and `ar-ary-arz-Arab`: it doesn’t check extended language subtags.

Lib/gftools/meta.py
```diff
@@ -0,0 +1,76 @@
...
+    if code.language and not VALIDITY.fullmatch(code.language):
+        log.warning(f"Invalid language code '{code.language}' in '{value}'")
+    if not VALIDITY.fullmatch(code.script):
+        log.warning(f"Invalid script code '{code.script}' in '{value}'")
+    if code.territory and not VALIDITY.fullmatch(code.territory):
+        log.warning(f"Invalid territory code '{code.territory}' in '{value}'")
+    if code.variants:
+        for variant in code.variants:
+            if not VALIDITY.fullmatch(variant):
+                log.warning(f"Invalid variant code '{variant}' in '{value}'")
```
dscorbett, line 75: If any warning is logged, this function should return false. Currently, it warns on `Cyrl-Serbia` but still returns true.

Lib/gftools/meta.py
```diff
@@ -0,0 +1,76 @@
...
+
+def _language_list_to_string(values):
+    return ", ".join(
+        sorted([value for value in set(values) if _validate_scriptlangtag(value)])
+    )
+
+
+def _validate_scriptlangtag(value):
+    try:
+        code = Language.get(value)
```
dscorbett, line 54: `Language.get` by default does some normalization of invalid inputs, as in `eng` to `en`. That means that all of the validation in this function is done in terms of the normalized code, whereas the caller of this function interprets the return value in terms of the original code. Either this function should return the normalized code so the caller can use it, or this line should be `code = Language.get(value, normalize=False)`.

dscorbett:
langcodes 3.2.1 has been released and includes better tag validation.

m4rc1e:
@simoncozens ready for review?

anthrotype:
I'm late to the party, but it would have been nice if this logic to build meta table were in ufo2ft (by reading from the public.openTypeMeta lib key)

anthrotype:
oops stupid me, we already do support that https://github.com/googlefonts/ufo2ft/blob/0cce3d76e5c901c0191750d9a036f4f6fc19174b/Lib/ufo2ft/outlineCompiler.py#L994 (sorry Denis!)

anthrotype:
I guess you want to be able to add a meta table directly to an existing binary font, right? Ideally that would be part of the source itself. If there's anything in the existing ufo2ft implementation that does not suit your needs we can change that.

simoncozens:
> I guess you want to be able to add a meta table directly to an existing binary font, right?

Actually we did want to add it to the font while building. I didn't know ufo2ft did this already. This is probably redundant then.

anthrotype:
> I didn't know ufo2ft did this already. 

hehe, even me I had forgotten 😄

m4rc1e:
What's the support like in glyphsapp/glyphsLib for this table?

I can already predict what you're all thinking 

image

anthrotype:
LOL

what about https://forum.glyphsapp.com/t/feature-request-support-for-public-opentypemeta/9039 ?

anthrotype:
> So the only useful part of this is validity checking for the dlng/slng tags. I'm happy to upstream that into ufo2ft if you'd want it, @anthrotype.

that'd be nice. How 'heavy' is the langcodes dependency?

simoncozens:
It doesn't have any required dependencies itself.

anthrotype:
sgtm

simoncozens:
I think this idea has died, and if it is to be picked up we'll do it in glyphsLib.

PR #403, opened 2021-08-19
Ensure that update_nametable correctly generates typo records

m4rc1e:
The current implementation will generate incorrect typographic name records for style names which do not include a weight token e.g:

family_name: Roboto
style_name: 8pt SemiCondensed

result:
familyName id1: Roboto 8pt SemiCondensed (correct since non wght and ital tokens are appended to the family name)
subFamilyname id2: Regular (correct)
typoFamilyName id16: Roboto 8pt SemiCondensed (not needed since it is the same as the familyName
typoSubFamilyName id17: "" (not needed since it is the same as the styleName. Should never be blank!)

My new heuristic will delete the typo records for the example given. The existing testcases also pass and I've added new ones to catch this issue.

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -364,24 +364,20 @@ def update_nametable(ttFont, family_name=None, style_name=None):
     if not style_name:
         style_name = font_stylename(ttFont)
 
-    is_ribbi = style_name in ("Regular", "Bold", "Italic", "Bold Italic")
+    ribbi = ("Regular", "Bold", "Italic", "Bold Italic")
+    tokens = family_name.split() + style_name.split()
 
-    nameids = {}
-    if is_ribbi:
-        nameids[1] = family_name
-        nameids[2] = style_name
-    else:
-        tokens = style_name.split()
-        family_name_suffix = " ".join([t for t in tokens if t not in ["Italic"]])
-        nameids[1] = f"{family_name} {family_name_suffix}".strip()
-        nameids[2] = "Regular" if "Italic" not in tokens else "Italic"
-
-        typo_family_suffix = " ".join(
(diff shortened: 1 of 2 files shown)


thlinard:
> typoFamilyName id16: Roboto 8pt SemiCondensed (not needed since it is the same as the familyName
> typoSubFamilyName id17: "" (not needed since it is the same as the styleName. Should never be blank!)

Isn't the very purpose of nameID 16 and 17 to overcome the RIBBI limitations of nameID 1 and 2 and to have:

typoFamilyName id16: Roboto 
typoSubFamilyName id17: 8pt SemiCondensed

Or at least, if you want to emulate current Adobe practices:
typoFamilyName id16: Roboto 8pt
typoSubFamilyName id17: SemiCondensed

?

m4rc1e:
@thlinard we append any token which isn't wght or ital to the family name for static fonts, hence we id16 is Roboto 8pt SemiCondensed.

thlinard:
> @thlinard we append any token which isn't wght or ital to the family name for static fonts, hence we id16 is Roboto 8pt SemiCondensed.

Microsoft created nameIDs 21 and 22 for the case where nameID 16 and 17 "do not conform to the WWS model. (That is, in case the entry for ID 17 includes qualifiers for some attribute other than weight, width or slope.) "

Wouldn't you want to request the creation of nameID 26 and 27 for names that do not conform to your model of weight + ital, rather than changing the definition of nameID 16 and 17?

m4rc1e:
We add non weight and italic tokens to the family name so that we maintain backwards compatibility with older families that we've already released e.g Cabin VF can instantiate both the Cabin and Cabin Condensed families we currently serve.

PR #410, opened 2021-09-08
Don't add dummy DSIG by default any more

simoncozens:
Fontbakery problem detection and gftools fixing has got out of sync again. (*as will always happen when the detection and the fixes are in different places...*) As per https://github.com/googlefonts/fontbakery/issues/3398, we don't want fonts to have a DSIG table:

> In fact, I think we should flip the check to WARN people if a DSIG is found at all, and tell them to remove it.


m4rc1e:
Thanks!

> as will always happen when the detection and the fixes are in different places...

Agreed. It is a pita.

PR #411, opened 2021-09-08
Drop Debg table on fix

simoncozens:
Guess what! It's another [fontbakery check](https://github.com/googlefonts/fontbakery/issues/3357) that got out of sync with the fix script!


m4rc1e:
🤣

PR #419, opened 2021-10-04
builder: add ability to compile vtt sources using vttLib

m4rc1e:
This PR will allow the builder to compile vtt hints using vttLib.

@simoncozens I'm deliberating whether we should write a python wrapper for the newly released vttCompiler and use this instead of vttLib? I get cross platform builds are still an ongoing issue though.


simoncozens:
The vttcompiler code is interesting. I'm looking at how much work a Rust port would be...

PR #420, opened 2021-10-05
gen-html: take local screenshots using selenium

m4rc1e:
I want us to move away from Browserstack since they haven't updated their screenshot browsers for many years. 

My end goal is to use a github actions matrix so we can take screenshots on different platforms. I have a prototype which is working nicely. I may try and roll it out to google/fonts this week if I have the time.

Once I've got this sorted, I will happily reaccept the MVAR table, https://github.com/googlefonts/fontbakery/issues/3413


m4rc1e:
Closing this since I've got something similar in diffenator2

PR #423, opened 2021-10-08
Gf glyphset Kernel and Latin Core

RosaWagner:
fixes #450 
When done, we can talk about the bigger "plus" sets.


Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
```diff
@@ -95,24 +96,18 @@
 0x007C | VERTICAL LINE
 0x007D } RIGHT CURLY BRACKET
 0x007E ~ TILDE
-0x00A0   NO-BREAK SPACE
 0x00A1 ¡ INVERTED EXCLAMATION MARK
 0x00A2 ¢ CENT SIGN
 0x00A3 £ POUND SIGN
-0x00A4 ¤ CURRENCY SIGN
-0x00A5 ¥ YEN SIGN
```
m4rc1e, line 103: I wouldn't remove the yen since it's heavily traded, https://en.wikipedia.org/wiki/Template:Most_traded_currencies. Please ignore if people represent it using other characters though.

Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
```diff
@@ -123,9 +118,6 @@
 0x00B9 ¹ SUPERSCRIPT ONE
 0x00BA º MASCULINE ORDINAL INDICATOR
 0x00BB » RIGHT-POINTING DOUBLE ANGLE QUOTATION MARK
-0x00BC ¼ VULGAR FRACTION ONE QUARTER
-0x00BD ½ VULGAR FRACTION ONE HALF
-0x00BE ¾ VULGAR FRACTION THREE QUARTERS
```
m4rc1e, line 128: I'd honestly keep these fractions. I know this is anecdotal but I use these fractions frequently and would be annoyed if a font didn't include them.

Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
```diff
@@ -1,6 +1,8 @@
-         NULL
+
+#$ include GF-latin-ascii_unique-glyphs.nam
+         .notdef
+         .null
 0x000D   CR
-0x0020   SPACE
```
vv-monsalve, line 3: 0020 `space` is required
00A0 `nbspace` was the only one to be removed

Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
vv-monsalve: We've talked about moving the fractions in the GF Plus encoding which is currently the basic set for commissioned fonts. 
Latin Core scope/goal would be to include the essential to support "most common" Latin languages as a basic ground for new fonts (particularly for not commissioned ones).

Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
simoncozens: I think it got moved to the top.

vv-monsalve: Didn't find it with `cmnd+f`, not by `SPACE` or `0020`

Lib/gftools/encodings/GF Glyph Sets/GF-latin-core_unique-glyphs.nam
vv-monsalve: It would take us back to one of the original concerns about including (or not) the `fraction` as part of the core set. Original reasoning around that was due to bad designed/spaced fractions as well as missing OT feature for them.

felipesanches:
@RosaWagner, since I merged #453, we will now need to migrate this PR to https://github.com/googlefonts/glyphsets/

And I can help with that, for sure!

RosaWagner:
Yep, I am redoing it to avoid conflict and separate the 2 glyphsets

PR #431, opened 2021-10-25
test on macos runner

m4rc1e:
Seems we're having some issues regarding pygit2 on OS X. Let's see if a dflt mac runner complains.


m4rc1e:
Will be getting an M1 on Thursday so I'll close this.

PR #436, opened 2021-11-08
builder: refactor so babelfont isn't needed

m4rc1e:
We currently only use babelfont to determine what the familyName should be. It seems tidier to just use glyphsLib, defcon etc in order to do this.

If we require babelfont in the future, I have no problems with readding it.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -97,7 +97,6 @@ required, all others have sensible defaults:
 
 """
 
-from babelfont import Babelfont
 from fontmake.font_project import FontProject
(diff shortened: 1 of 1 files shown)


simoncozens:
Strongly approve. Babelfont 3 is going to look very different from Babelfont 2 so getting rid of it now is a good choice.

m4rc1e:
@simoncozens I've quickly checked my implementation against ufos, glyphs and designspaces by doing, `gftools builder $src`. All of them could correctly infer the familyName (even checked with pdb). You happy for me to merge?

simoncozens:
Sounds good, yes.

PR #438, opened 2021-11-10
Create standalone executables for popular tools

### LINKED ISSUE #3: Create setup.py to enable pip packaging/installation
Would be good to have a setup.py script, so we could install this repo's tools via pip

m4rc1e:
If I was a type designer, I'd have no inclination to learn about dependency management, installing package managers and ensuring i have the correct libs etc. I want an easy life. A life of drawing black and white shapes and admiring beautiful whitespace. Well that time is nigh. Introducing standalone executable for gftools builder and gftools qa.

Still draft since I want to get these working cross platform. They will also be auto genned and signed for each release using github actions.

fixes #3


m4rc1e:
Getting there slowly. Linux and Win execs can now be built and attached to the release.

Screenshot 2021-11-19 at 12 04 27

PR #444, opened 2021-11-26
Warn M1 users about pip upgrade

simoncozens:


README.md
```diff
@@ -54,11 +54,13 @@ Please install these tools using pip:
 
     pip install gftools
 
+(Mac OS users on M1 chips will need to upgrade pip first to support universal wheels: `pip install -u pip`.)
```
felipesanches, line 57: Is it lowercase `-u` on Mac?

I'm only asking this because, at least on GNU+Linux, there's no `-u`. The correct way is uppercase `-U` and I would expect the flag to be the same on all systems.

simoncozens: Oops, you're right. I was doing it from memory!

README.md
```diff
@@ -54,11 +54,13 @@ Please install these tools using pip:
 
     pip install gftools
 
+(Mac OS users on M1 chips will need to upgrade pip first to support universal wheels: `pip install -U pip`.)
```
anthrotype, line 57: doesn't it depend on where they get their python from, though? If one installs, say, 3.10 from Homebrew or Python.org, I bet they do have pip up to date. Perhaps it's just the python that 's available from Apple that is out of date? Just speculating as I am not on M1 yet.

simoncozens: You're probably technically right, but the point is to give people a nice clear instruction to head off problems later. There's no real harm in people upgrading pip if they're using a non-Apple Python, and it makes the instructions easier to follow if we say Just Do It.

anthrotype: yeah, upgrading pip is the first thing I always do myself (along with setuptools and wheel) whenever I create a new venv

m4rc1e:
Thanks Simon!

PR #445, opened 2021-11-28
Require pandas at runtime only

simoncozens:
Installing gftools on M1 is obnoxiously slow because there isn't a wheel for pandas and it has to be built from scratch for every venv you install it into, and it takes *ages*. And you don't even need it 90% of the time! It's only used if you're using `gftools add-designer` *and* your input is an Excel spreadsheet.

Now that we're using the gftools-builder in UFR, and hence gftools is migrating towards more of a general userbase kind of thing, we need to be more aware of this kind of stuff.

This PR removes the pandas dependency, but tells you off if you do `gftools add-designer --spreadsheet` without pandas installed.

### DIFF
--- a/bin/gftools-add-designer.py
+++ b/bin/gftools-add-designer.py
@@ -33,7 +33,6 @@ import os
 from PIL import Image
 from gftools.designers_pb2 import DesignerInfoProto
 from google.protobuf import text_format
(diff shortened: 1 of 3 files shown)


m4rc1e:
Thanks. This is much better.

PR #446, opened 2021-12-02
Add fix_glyph_names()

simoncozens:
This adds a fix for the fontbakery glyph naming check; it uses ufo2ft's logic to rename glyphs to production names. This is useful because in a variable-first workflow, you build the variable font first and then add GSUB/GPOS/GDEF tables to it from a variable feature file afterwards. To do this, the glyph names in the font have to be the same as the ones in the feature file, and hence you have to pass `--no-production-names` to fontmake, and then do the renaming afterwards. (You could use Rust's `ttf-rename-glyphs` but that's not part of our workflow yet. ;-)

Because this is a slightly obscure use case, I've not added it to `fix_font`.

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -8,6 +8,7 @@ from fontTools.ttLib import TTFont, newTable, getTableModule
 from fontTools.ttLib.tables import ttProgram
 from fontTools.ttLib.tables._c_m_a_p import CmapSubtable
 from fontTools.ttLib.tables._f_v_a_r import NamedInstance
+from ufo2ft.postProcessor import PostProcessor
 from gftools.util.google_fonts import _KNOWN_WEIGHTS
 from gftools.utils import (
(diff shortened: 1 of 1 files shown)


simoncozens:
This is clever but doesn't work. If you decompile the cmap table (as we did to find the existing Unicode mappings), the mappings are stored internally in the font object as codepoint => glyph names. Changing the glyph names in the `TTFont` object doesn't change the names in the cmap table objects.

(Incidentally this is why fonttools-rs uses glyph IDs rather than glyph names everywhere.)

m4rc1e:
I probably being dense here but I thought all our VFs already have production names? do you need this because you have a custom build pipeline that only uses a few pieces of fontmake?

simoncozens:
Yeah, basically. I've put it into a custom script in my font's repo, but I figured if this is something that fontbakery has a check for, it should also be something that gftools should have a fix for...

PR #447, opened 2021-12-03
push-stats: add commit table

m4rc1e:
Include a table which contains info for every commit.


m4rc1e:
Screenshot 2021-12-03 at 14 53 29

PR #448, opened 2021-12-08
gen-push-lists added

m4rc1e:
This tool generates the [to_production.txt](https://github.com/google/fonts/blob/main/to_production.txt) and [to_sandbox.txt](https://github.com/google/fonts/blob/main/to_sandbox.txt) files based on the card labels in the [traffic jam](https://github.com/google/fonts/projects/1).

@RosaWagner Could you quickly test this?


RosaWagner:
It looks like it works like a charm. I got the list with no error, now I gonna check if all is correct.

PR #450, opened 2021-12-10
Removed deprecated signs in GF Latin Core

RosaWagner:
While we are trying to redefine what should be GF Latin Core, we agreed on removing deprecated signs. Indeed, Fontbakery fails if fonts don't match this glyphset, although it contains glyphs that are not necessary or in use anymore. This PR complete the 1st step of redefining the glyphset, see #423 for the complete proposal in progress.


davelab6 (approved):

davelab6:
@RosaWagner who do you want to review/merge this?

RosaWagner:
you haha, so my name is not the only one on this

davelab6:
I chatted with @rsheeter today about this

Since this is part of the glyphs sets, not the subset definitions, this is fine to merge; but one follow on could be to update the latin.nam used in the API. 

Rod asked if we do want to do that, first someone (Rosalie, Marc, Simon, Felipe, etc) ought run an/some experiments on the TTFs in the github.com/google/fonts repo, to better understand the impact of this change. 

That is, run the fonttools subsetter over the TTFs with both current and proposed latin.nam subsets, and tabulate and average the differences, then woff2 them both and tabulate and average those differences too, so we can know what the filesize reduction will be in the average case, for the top 5, 10 families, etc. 

Also, Rod pointed out that the removed characters will likely move to the latin-ext.nam set, or a new set; the additional API transfers needed by users who do use the rare characters might be a net loss if latin-ext has a bunch of other stuff in it they don't use; and a new set wouldn't kern/shape with other sets, and for just these 3 characters (soft hyphen, division slash and ¤) a new set wouldn't make sense.

PR #451, opened 2021-12-16
Support multiple categories

rsheeter:
Note that the LAST category will be used by fonts.google.com [for now]. This is meant to unblock data collection for multiple categories.


davelab6:
@RosaWagner I think this is what you want :D

RosaWagner:
Indeed, so I close your draft PR #4141 that doesn't give the correct example. For full clarity we should have, for example, in `metadata.pb`:

```
category: "SANS_SERIF" -> second category above, not yet supported 
category: "MONOSPACE" -> first category under, currently in use
```

davelab6:
Updated my PR. @m4rc1e is this good to merge?

m4rc1e:
When I run add-font on an existing family, I get the following traceback:

```
  File "/Users/marcfoley/Type/tools/bin/gftools-add-font.py", line 271, in <module>
    app.run(main)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.9/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.9/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "/Users/marcfoley/Type/tools/bin/gftools-add-font.py", line 257, in main
    metadata = _MakeMetadata(fontdir, is_new)
  File "/Users/marcfoley/Type/tools/bin/gftools-add-font.py", line 134, in _MakeMetadata
    metadata.category = old_metadata.category
  File "/Users/marcfoley/Type/tools/venv/lib/python3.9/site-packages/google/protobuf/internal/python_message.py", line 664, in setter
    raise AttributeError('Assignment not allowed to repeated field '
AttributeError: Assignment not allowed to repeated field "category" in protocol message object.
```

`metadata.category = old_metadata.category` is the issue here. I've added a commit which simply appends the old categories instead. I've tested this solution on both new and old families and it works ok.

RosaWagner:
Would it be supported by packager then? Like we just add a category in the yaml

m4rc1e:
Good point. I'll take a look.

m4rc1e:
For the packager, we need to decide how we're going to input the category. We have two options that I can think of.

1. Require the yaml file to use a list for the category e.g

```YAML
category:
  - DISPLAY
  - SERIF
...

Or still keep it as a string but use a comma as a delimiter e.g:

```YAML
category: DISPLAY, SERIF
```

Option 2 may be easier but option 1 seems more legit.

RosaWagner:
Since we use comma for designers, should be the same for categories to avoid confusing users; the output will be different but this not the concern of users.

graphicore:
I'm looking at the `strityaml` docs to figure out how to do this, as comma separated string is basically the wrong data type, we want a list, to match the nature of the `repeated` entry of `fonts_public.proto`. So, good news, there's a [`CommaSeparated`](https://hitchdev.com/strictyaml/using/alpha/scalar/comma-separated/) type.

A thing I'd personally would prefer would be to have an ordered set, i.e. a list with only unique entries, and there's also a type: [`UniqueSeq`](https://hitchdev.com/strictyaml/using/alpha/compound/sequences-of-unique-items/)

The `CommaSeparated` would be specified something like: `CommaSeparated(Enum(CATEGORIES))` in the YAML written as a comma separated string.

And the `UniqueSeq` would be something like: `UniqueSeq(Enum(CATEGORIES))` in the YAML written as a list.

I wonder if: `UniqueSeq(CommaSeparated(Enum(CATEGORIES)))` would work as well. 

I'm going to try the above and report back which options work....

What works:

- `CommaSeparated(Enum(CATEGORIES))` input as string
- `Seq(Enum(CATEGORIES))` input as list
- `UniqueSeq(Enum(CATEGORIES))` input as list

What Fails:

- `UniqueSeq(CommaSeparated(Enum(CATEGORIES)))` input as string. This does not combine that way.

m4rc1e:
Welcome back everyone. I've decided to use the UniqueSeq approach. I've chosen not to parse a string ourselves since a UniqueSeq gives us auto validation. However, this now means we'll need to declare the category in the following manner:


```YAML
category:
- DISPLAY
- SANS_SERIF
```

I've updated the template.yml so it;s clear that users now need to do this:

```YAML
# Choose multiple of: DISPLAY, SERIF, SANS_SERIF, HANDWRITING, MONOSPACE. Currently, the last item will be used
# by Google Fonts
category:
  # - SANS_SERIF
  # - SERIF
  # - DISPLAY
  # - MONOSPACE
  # - HANDWRITING
```


@graphicore could you take a quick look?

Lib/gftools/packager.py
```diff
@@ -671,7 +674,10 @@ def _upstream_conf_from_yaml_metadata(
                                        , allow_flow_style=True)
   for k,v in upstream_conf.items():
     if v is None: continue
-    upstream_conf_yaml[k] = v
+    if k == "category":
+      upstream_conf_yaml[k] = v._values
```
m4rc1e, line 678: We cannot simply do `upstream_conf_yaml[k] = v` for the category since it's not a list.

Lib/gftools/packager.py
```diff
@@ -914,7 +920,12 @@ def _create_or_update_metadata_pb(upstream_conf: YAML,
   for font in metadata.fonts:
     font.name = upstream_conf['name']
   metadata.designer = upstream_conf['designer']
-  metadata.category = upstream_conf['category']
+
+  while metadata.category:
+    metadata.category.pop()
```
m4rc1e, line 925: Would be nice if there was a method to clear this container but it doesn't [exist](https://googleapis.dev/python/protobuf/latest/google/protobuf/internal/containers.html#google.protobuf.internal.containers.RepeatedScalarFieldContainer) hence popping till empty.

graphicore (changes requested):
Ok, I tested this and it looks good to me.

I was a bit concerned about the case when old `upstream.yaml` files have a category set as a string, but I think the handling works good.

In the fonts repo, there are two `upstream.yaml` files which do this. These seem to be an old style, when redundancies with `METADATA.pb` and comments where not removed. Good for testing:

```
$ find fonts/ -type f -name upstream.yaml | xargs grep category
fonts/ofl/chilanka/upstream.yaml:category: HANDWRITING
fonts/ofl/literata/upstream.yaml:category: SERIF
```

I was pleased to see this reported:

```
$ gftools packager literata packager_prep/
Font Family "literata" is on Google Fonts under "ofl".
Using upstream.yaml from google/fonts for literata.
The configuration has schema errors:

when expecting a unique sequence
  in "<unicode string>", line 18, column 1:
    category: SERIF
    ^ (line: 18)
found arbitrary text
  in "<unicode string>", line 20, column 1:
    # Full name of the type designer ... 
    ^ (line: 20)
Your options e=edit,[q]=quit program:
```

Using the `e` option I was able to fix the issue by changing the category entry into a list.

The only thing that comes to my mind is the question whether that case should be detected and handled automatically. But, I don't think it will come up often, as new `upstream.yaml` files won't have this issue, and existing ones in the fonts repo that have a category defined are just these two (chilanka and literata).

Lib/gftools/packager.py
graphicore: All of the following work ([via](https://developers.google.com/protocol-buffers/docs/reference/python-generated#fields)):

```py
metadata.category[:] = []
metadata.ClearField('category')
del metadata.category[:]
```

Lib/gftools/packager.py
```diff
@@ -914,7 +920,12 @@ def _create_or_update_metadata_pb(upstream_conf: YAML,
...
+  for cat in upstream_conf['category']:
+    metadata.category.append(cat)
```
graphicore, line 927: Actually, from the insight of my comment before, this is sufficient to clear the field and to assign the new values:

```py
metadata.category[:] = upstream_conf['category']
```

graphicore (changes requested):

Lib/gftools/packager.py
graphicore: I can't actually reproduce the issue, for me it works when `category` is a `<class 'google.protobuf.pyext._message.RepeatedScalarContainer'>`.

Maybe this is an issue with the version of the `protobuf` package?

Despite of that, maybe it would be nicer to do this before in line 654:

```py
   upstream_conf.update({
      'designer': metadata.designer or None,
      'category': list(metadata.category) or None,
      'name': metadata.name  or None,
      # we won't get this just now in most cases!
      'repository_url': metadata.source.repository_url or None,
    })
```

graphicore (changes requested):

graphicore:
@m4rc1e Sorry for the late reply, I'm just since yesterday back on keyboard. I submitted the review, I hope its not too confusing with the corrections I made on my own suggestions.

Lib/gftools/packager.py
m4rc1e: I'm using protobuf 3.19.1 which seems to be the latest. I'm able to replicate it. I'll use your last proposal since that is definitely the cleanest. Thank you.

m4rc1e:
@graphicore Thank you for the thorough review. It wasn't hard at all to follow! I've implemented your feedback so ptal.

Lib/gftools/packager.py
m4rc1e: Thanks. I've opted for ClearField.

Lib/gftools/packager.py
```diff
@@ -921,10 +918,8 @@ def _create_or_update_metadata_pb(upstream_conf: YAML,
     font.name = upstream_conf['name']
   metadata.designer = upstream_conf['designer']
 
-  while metadata.category:
-    metadata.category.pop()
-  for cat in upstream_conf['category']:
-    metadata.category.append(cat)
+  metadata.ClearField('category')
+  metadata.category[:] = upstream_conf['category']
```
graphicore, line 922: The `ClearField` line is unnecessary because the following line reassigns the entire list, it's not an append.

```py
metadata.category[:] = upstream_conf['category']
```

m4rc1e: Thanks. Updated.

graphicore (approved):

PR #453, opened 2021-12-17
use `glyphsets` python module

felipesanches:
Below are the most important changes from each release of the new `glyphsets` python module:

 ## 0.2.1 (2021-Dec-17)
  - Quick bugfix: undefined var `enc_path` (should be `nam_dir` instead!)

 ## 0.2.0 (2021-Dec-17)
 ### Noteworthy code-changes
  - Added function `set_encoding_path` that allows one to use `.nam` files from a different directory, other than the files shipped with the `glyphsets` module.
  - Minor code cleanup
  - This release includes a few tweaks needed to support usage in `gftools`.

 ## 0.1.0 (2021-Dec-14)
 ### Release notes
  - Initial release of the `glyphsets` python module.
  - Most of the code was migrated from the `gftools` repository (https://github.com/googlefonts/gftools/) so that glyphset data can be easily available to all our tools without having to also get the large dependency tree of `gftools`. The most immediate user of this module is `Font Bakery`, which needs to validate conformance of the Google Fonts glyphsets on font binaries being checked. (see https://github.com/googlefonts/fontbakery/issues/3533)
  - The second obvious user of this `glyphsets` module will be `gftools` itself. I'll be sending a pull request soon. All GFonts glyphset definitions will then be defined here on the `glyphsets` module, to avoid data duplication and to guarantee uniformity across tools.


felipesanches:
**Notes:**

These are the public functions from `gftools.util.google_fonts` that were migrated to `glyphsets.codepoints`:

```python
def CodepointFiles(subset):
def CodepointsInSubset(subset, unique_glyphs=False):
def CodepointsInFont(font_filename):
def CodepointsInNamelist(nam_filename, unique_glyphs=False, cache=None):
def GetCodepointFromLine(line):
def ParseNamelist(filename):
def ReadNameList(nam_filename, unique_glyphs=False, cache=None):
def SubsetForCodepoint(cp):
def SubsetsForCodepoint(cp):
def SubsetsInFont(file_path, min_pct, ext_min_pct=None):
```

These are private functions also migrated:
```python
def _LoadNamelistIncludes(item, unique_glyphs, cache):
def _ParseNamelist(lines):
def _ParseNamelistHeader(lines):
def _ReadNameList(cache, filename, unique_glyphs):
def _ReadNameListSafetyLayer(currently_including, cache, nam_filename,
                             unique_glyphs):
```

And during development, I crafted this, but ended up leaving it out in this PR, since I could not find any user of the feature:
```python
# FIXME: I think nobody actually uses the --nam_dir flag,
#        so we might as well drop this soon, unless someone
#        still wants to keep this private wrapper:
from absl import flags
FLAGS = flags.FLAGS
flags.DEFINE_string('nam_dir', None, 'nam file dir')
def _CodepointFileForSubset(subset):
  """This is a private wrapper to the CodepointFileForSubset function provided
  by the glyphsets module with the purpose of handling the nam_dir flag used in gftools.
  """
  from glyphsets.codepoints import (set_encoding_path,
                                    CodepointFileForSubset)

  if FLAGS.nam_dir:
    # expanduser so we can do things like --nam_dir=~/oss/googlefontdirectory/
    enc_path = os.path.expanduser(FLAGS.nam_dir)
    if not os.path.exists(enc_path):
      raise OSError(errno.ENOTDIR, 'No such directory', enc_path)

    set_encoding_path(enc_path)

  return CodepointFileForSubset(subset)
```

felipesanches:
I was concerned that this PR could be troublesome to the internal Google usage of gftools, but @davelab6 gave me the green light to implement these changes as the internal fork is not anymore kept in sync with the public one.

All of the code-tests have passed and I am happy with the results, so I think it is safe to now merge this.

@vv-monsalve, @m4rc1e, @RosaWagner, @simoncozens and others, from now on, please submit any PR related to glyph set updates directly to the `glyphsets` repo at https://github.com/googlefonts/glyphsets

Some open PRs in this repo (such as #450) will need to be updated/migrated and I can help with that.

Issue #456, opened 2022-01-05
Include gftools-lang-support in gftools-add-font


m4rc1e:
We had a commit which added language fields to the METADATA.pb files in google/fonts, https://github.com/google/fonts/commit/633ebadbf5280a439a49ee831ab9446093b27f62.

If we plan on adding these fields to every metadata file, wouldn't it just be easier to include this functionality in the existing gftools add-font tool? atm users need to run both add-font and lang-support.

@nathan-williams wdyt? I'll happily implement this if you want as well.

discovered in https://github.com/google/fonts/pull/4194

RosaWagner:
Ah, lang-support should be also run on new families like… by us?

m4rc1e:
Fixed in #460 

@RosaWagner add-font now does this automatically. No need to do anything.

PR #460, opened 2022-01-07
wip add-font: add languages

m4rc1e:
I spoke to @nathan-williams regarding #456 and we decided on the following:

- Move Language Support functions to `gftools.util.google_fonts`
- Deprecate `gftools lang-support` (I now think we should keep it)
- add language generation to `gftools add-font`

This PR does the above by:

Adding the [google/fonts lang dir](https://github.com/google/fonts/tree/main/lang) as package data. Since we now have a fixed home for this data, we don't need to declare the lang flag in either add-font or lang-support. However, users can still declare the flag so we don't break backwards compatibility or so they can use their own custom package data.

Breaking apart `AddLanguageSupportMetadata`. This function wasn't very reusable so I dropped it and improved the api of its sibling functions. Currently, some function inputs take a path and then load a TTFont object. I've opted to drop loading,saving etc in each func. It's much easier to just accept TTFont etc as input from the start.

---

The changes I've made means we can keep `lang-support`. I now see why this script was created. I can imagine the brief was something along the lines of "Hey Nate, can you add language fields to each metadata file? This seems like quite a useful batch script so I've decided to keep it.

@nathan-williams due to the amount of changes, it may be easier to review this by looking at each commit. I've kept each commit quite clean for this purpose. Would you also like me to extend this tidy up to the other lang scripts? Atm, I've only touched add-font and lang-support.


m4rc1e (commented):
@nathan-williams I added these fields myself so they may be incorrect. It would be great if you could track down the correct fonts_public.proto file and I can use this instead.

m4rc1e:
@rsheeter is it ok to get Nate to review this?

rsheeter:
We should prefer to edit the proto file internally and then publish here. Editing here gets us out of sync. @nathan-williams could likely do that since he knows all about the lang fields.

rsheeter:
It's tiresome to review 1k+ files; suggest splitting the manual changes and the automated changes into their own PRs

m4rc1e:
> We should prefer to edit the proto file internally and then publish here. Editing here gets us out of sync. @nathan-williams could likely do that since he knows all about the lang fields.

Agreed. I've just removed this commit. If he could commit the latest fonts_public.proto file and compile it to this PR, we should be good.

> It's tiresome to review 1k+ files; suggest splitting the manual changes and the automated changes into their own PRs

Ok. nam files have been moved to #461

Lib/gftools/util/google_fonts.py
```diff
@@ -459,3 +462,85 @@ def LicenseFromPath(path):
...
+  for textproto_file in glob.iglob(os.path.join(regions_dir, '*.textproto')):
+    with open(textproto_file, 'r', encoding='utf-8') as f:
+      region = text_format.Parse(f.read(), fonts_pb2.RegionProto())
+      regions[region.id] = region
+  return regions
+
+
+def SupportedLanguages(
+  ttFont,
+  languages=LoadLanguages(resource_filename("gftools", "lang/languages"))
```
nathan-williams, line 495: @m4rc1e are you planning to copy the lang metadata into gftools? I'm worried about the two views of lang metadata falling out of sync.

Lib/gftools/util/google_fonts.py
```diff
@@ -459,3 +462,85 @@ def LicenseFromPath(path):
...
+def GetExemplarFont(family):
+  assert len(family.fonts) > 0, 'Unable to select exemplar in family with no fonts: ' + family.name
+  for font in family.fonts:
+    if font.style == 'normal' and font.weight == 400:
+      # Prefer default style (Regular, not Italic)
+      return font
+  return family.fonts[0]
+
+
+def language_comments(languages):
```
nathan-williams, line 526: For consistency, this function should probably be named something closer to LanguageComments

nathan-williams: Also, can you add a comment explaining what this function is for? Now that the function lives over here, I don't think its purpose is obvious.

bin/gftools-add-font.py
```diff
@@ -255,18 +268,18 @@ def main(argv):
   if os.path.isfile(old_metadata_file):
     is_new = False
 
+  language_comments = fonts.language_comments(
+    fonts.LoadLanguages(os.path.join(FLAGS.lang, 'languages'))
```
nathan-williams, line 272: Happy to see the lang flag still present. Thanks!

nathan-williams (approved):

Lib/gftools/util/google_fonts.py
m4rc1e: For the time being yes.

We currently do this for the axis registry as well. In order to keep everything in sync, we have a [script](https://github.com/googlefonts/gftools/blob/main/update_reg.sh) which will download the relevant folder from google/fonts and add it here. We can do the same for the lang dir.

Ideally, we should have all of this stuff in a single repo that we can import into google/fonts, googlefonts/tools and fontbakery.

PR #470, opened 2022-01-27
update axis metedata

zhaoxiong0211:
introduce a new field, display_name, in fallback and a new filed, ui_type, in axes


m4rc1e:
Just a heads up but we also have an axis registry in https://github.com/google/fonts. This registry is merely a copy of it. I'll leave it up to Nate though.

PR #471, opened 2022-01-28
Detect script and pass to ttfautohint

simoncozens:
The Noto build scripts pass a `-D<script>` argument to ttfautohint to enable script-specific hinting adjustments. This is a nice idea, and we should do it if possible. Also, the way Noto implements it is [horrible and arbitrary](https://github.com/googlefonts/noto-source/blob/3eab4dae98b0a40d38d216cab2a66b32000f3bfc/misc_scripts/publish-one-noto-font.sh#L48) and I'd like to move Noto's build scripts to cleaner code based on gftools where possible.

This checks whether a font file has a clear "winning" script (defined as having more than twice as many code points as any other script in the font) and, if ttfautohint has a script-specific handler for that script, passes the appropriate `-D` flag.


m4rc1e:
Just a thought, why not include a bool option in the config file for this? could be something like `ttfaUseScript`?

simoncozens:
Yeah, that'll work for me. Will do.

m4rc1e:
Thanks Simon. I've just added `ttfaUseScript` to the config file schema.

Issue #475, opened 2022-01-31
fs_selection WWS bit should be 0 if no name ID 21 and 22


moontypespace:
Hi,

according to Microsofts OT Spec the fsSelection bit 8 (=WWS) should be set if name ID 21 and 22 are NOT used.

_Many typographic families contains faces that differ only in one or more of the attributes weight, width and slope. Even though a family might have a large number of member faces, if the variations are in these attributes only, then family and subfamily names provided in the 'name' table using IDs 1 and 2 or 16 and 17 will be consistent with a weight/width/slope family model. In this case, bit 8 should be set, and 'name' entries for name IDs 21 and 22 should not be included._
https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection

Currently, the following code does not take care of this:
https://github.com/googlefonts/gftools/blob/f0d02b32af0ee57c5df2b9ca683b071a9c029a44/Lib/gftools/fix.py#L224

Cheers, 
Olli

moontypespace:
It should include something like:
```
    name_table_obj = font_obj['name']
    name_id_21 = name_table_obj.getDebugName(21)
    name_id_22 = name_table_obj.getDebugName(22)

    if name_id_21 is None and name_id_22 is None:
        fs_selection = font_obj["OS/2"].fsSelection
        fs_selection |= 1 << 8
        font_obj["OS/2"].fsSelection = fs_selection
```

m4rc1e:
Thanks for both the issue and PR.

Are you planning on using gftools to release families which may have Condensed/Expanded styles etc? I'm afraid atm, this isn't possible.

I'd be interested to hear your requirements since I'd like us to make the tools in this repo more flexible.

moontypespace:
@m4rc1e Thanks for your patient feedback. As mentioned in the PR, it seems my understanding of the use of the wws bit differs a bit from yours. Even thou I 'separate' the family into their 'wdiths', I set the wws bit when I don't use name ID 21 and 22.

m4rc1e:
Our name records and bit settings are specific to Google Fonts. They may not follow best practices. However, please note that Source Sans, Source Serif etc also do not contain either the WWS bit or nameIds 21,22. The only families we serve which do have the WWS bit enabled are Noto since they also contain width variants so it does fulfil the WWS model.

@twardoch I seem to remember you may have had a hand in deciding name records and bit settings?

m4rc1e:
[...]

PR #476, opened 2022-01-31
extend fix_fs_selection with wws bit fix

moontypespace:
Please see: https://github.com/googlefonts/gftools/issues/475

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -230,6 +230,12 @@ def fix_fs_selection(ttFont):
     # enable Regular bit for all other styles
     if not tokens & set(["Bold", "Italic"]):
         fs_selection |= 1 << 6
+
+    # set WWS bit if no name ID 21 and 22
+    name = ttFont['name']
+    if name.getDebugName(21) is None and name.getDebugName(22) is None:
+        fs_selection |= 1 << 8
+
     ttFont["OS/2"].fsSelection = fs_selection
     return old_selection != fs_selection
 


m4rc1e:
If we do eventually accept families with differing widths, we'll reopen this.

moontypespace:
> Thank you for your PR.
> 
> Our fonts do not conform to the WWS model since we release styles which don't have a normal width as new families, https://fonts.google.com/?query=condensed, hence why we don't set this bit.

@m4rc1e Maybe there is a misunderstanding. But as far as I understand the situation your example of 'Roboto Condensed' fits to WWS (Weight/Width/Slope): 'Roboto Condensed Bold Italic' -> weight= Bold, width=Condensed, slope=Italic.

Do I overlook something?

m4rc1e:
On fonts.google.com, we release Roboto and Roboto Condensed as seperate families.

Roboto

```XML
<namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Roboto
    </namerecord>
```

Roboto Condensed
```XML
    <namerecord nameID="1" platformID="3" platEncID="1" langID="0x409">
      Roboto Condensed
    </namerecord>
```

moontypespace:
Agree @m4rc1e : 
Roboto Normal width name ID 16 = 'Roboto'
Roboto Condensed width name ID 16 = 'Roboto Consensed'

No name ID 21 and 22 in all fonts.

So, my understanding is, that in that case the wws bit should be set (if OS/2 table version 4 is used).

The Micosoft page also says:
_Thus, if a font has a version 4 or later OS/2 table, bit 8 should be set if and only if 'name' entries for IDs 16 and 17 are consistent with the WWS model and entries for IDs 21 and 22 are not included. Conversely, if bit 8 is not set, that will be interpreted to mean that the names provided by IDs 16 and 17 are not consistent with the WWS model and that 'name' entries for IDs 21 and 22 are included._

https://docs.microsoft.com/en-us/typography/opentype/spec/os2#:~:text=Thus%2C%20if%20a,22%20are%20included.

PR #480, opened 2022-02-01
add-designer: asciify filenames

### LINKED ISSUE #479: add-designer should use ascii only for png file name
cf https://github.com/google/fonts/issues/4262

m4rc1e:
Fixes #479.

@RosaWagner could you test this for me?

### DIFF
--- a/bin/gftools-add-designer.py
+++ b/bin/gftools-add-designer.py
@@ -30,6 +30,7 @@ $ gftools add-designer path/to/local/clone/fonts/catalog/designers "Theo Salvado
 import argparse
 from glob import glob
 import os
+from unidecode import unidecode
 from PIL import Image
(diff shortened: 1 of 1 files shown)


RosaWagner:
It works on Mirko's case

RosaWagner:
Also just tested "Großäî ÉMEæRÂÖLd ùbœfü" and return "grossaiemeaeraolduboefu", so it looks correct. I think the germand transfor `ü` to `ue`, but I don't think they care about our directory names system.

m4rc1e:
Could you also test using `--img_path` with an accented img filename?

RosaWagner:
image path gets renamed after the dir name

m4rc1e:
Cool. I'm happy with this,

PR #484, opened 2022-02-07
Two builder enhancements

simoncozens:
* Keep a list of created files - useful for zipping the output, checking into git, etc.
* Allow variable font builds to fail without killing the whole build.


m4rc1e:
Thanks Simon!

PR #488, opened 2022-02-09
Update gftools-qa to log INFO message when using fontbakery

### LINKED ISSUE #483: [gftools qa] Use -l INFO to stop skipping them
Important info are skip if we only show warn. Felipe just convinced us that we could benefit from displaying also the Info checks.

RosaWagner:
fixes #483

### DIFF
--- a/bin/gftools-qa.py
+++ b/bin/gftools-qa.py
@@ -186,7 +186,7 @@ class FontQA:
         out = os.path.join(self.out, "Fontbakery")
         mkdir(out)
         cmd = (
-            ["fontbakery", "check-googlefonts", "-l", "WARN"]
+            ["fontbakery", "check-googlefonts", "-l", "INFO"]
             + [f.reader.file.name for f in self.fonts]
             + ["-C"]
             + ["--ghmarkdown", os.path.join(out, "report.md")]


m4rc1e:
Unfortunately, this may make some FB reports too long for the gf bot to post as a comment (Github only allows comments under a certain length). However, I think we may have made the reports more succinct so this may not be an issue. If we notice it's an issue, we'll revert this.

Thanks!

RosaWagner:
we could use the --succinct argument too to shorten it?

RosaWagner:
Ah no never mind, it only compacts text in the terminal not in the report.

m4rc1e:
All good. Let's see how it goes.

felipesanches:
> we could use the --succinct argument too to shorten it?
> Ah no never mind, it only compacts text in the terminal not in the report.

Well... if that's the case, then we should consider making `--succinct` also affect the GitHub Markdown report output.

felipesanches:
> If we notice it's an issue, we'll revert this.

If there's an issue, then we can try to make the INFO reports shorter (via `--succinct` perhaps) instead of completely reverting this.

Please file a fontbakery issue if further changes are indeed needed.

PR #491, opened 2022-02-10
Add tool to check ufo MM compatibility

m4rc1e:
This pr includes a tool to check whether a bunch of ufos are MM compatible. Users are also able to select which ufo they want to compare the rest against.

I'm currently dealing with a VF which has 71 masters so this script is rather helpful to me. Imo the output is a touch easier to understand than the fontmake errors at this scale.

This script is rather primitive (written in around 40 mins) so don't expect fancy output atm. If there's more demand, I'll improve it.

### DIFF
--- a/bin/gftools-ufo-check-compatibility.py
+++ b/bin/gftools-ufo-check-compatibility.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""
+Check ufo sources are MM compatible. Very primitive.
+
+Usage:
+
+Basic:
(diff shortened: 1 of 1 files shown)


simoncozens:
This looks good, but you can probably replace most of the code with

```
from fontmake.compatibility import CompatibilityChecker

CompatibilityChecker(fonts).check()
```

m4rc1e:
Going to close this since the CompatiblityChecker output has improved.

PR #492, opened 2022-02-11
Use --succinct flag to FB

simoncozens:
See #488. In fact, fontbakery *does* remove the rationales when `--succinct` is passed (in reports, not just in terminal), and also we are seeing problems with reports failing to upload because they are too big.


simoncozens:
Huh. I was sure that `--succinct`  dropped rationales, but now I can't seem to get it to do so. Oh well, it will soon...

simoncozens:
Ah, it does in console and HTML but not currently --ghmarkdown. I'll fix it.

m4rc1e:
Cool! thank you!

PR #496, opened 2022-02-17
packager: rm --no-source flag

### LINKED ISSUE #494: packager: --no-source strips out language comments
@yanone noticed that language comments are stripped out if the packager is called with `--no-source`. I'm guessing the issue is in https://github.com/googlefonts/gftools/blob/main/Lib/gftools/packager.py#L1461-L1493.

I will get this fixed by the end of the day.

Also, we don't need to call the packager any more with `--no-source` because the source field has been added to the protobuf file, https://github.com/googlefonts/gftools/blob/main/Lib/gftools/fonts_public.proto#L23

m4rc1e:
This PR removes the `--no-source` flag since our METADATA.pb files now accept the source field.

This also fixes #494

### DIFF
--- a/Lib/gftools/packager.py
+++ b/Lib/gftools/packager.py
@@ -897,8 +897,7 @@ def _copy_upstream_files_from_dir(source_dir: str, files: dict,
 
 def _create_or_update_metadata_pb(upstream_conf: YAML,
                                   tmp_package_family_dir:str,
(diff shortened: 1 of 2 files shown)


m4rc1e:
The docs don't mention it and I'm fairly sure I've removed every reference.

PR #500, opened 2022-02-23
ufo-setter: added

m4rc1e:
Set attributes for a collection of ufos from the cli.


m4rc1e:
Decided to write this because I often write a load of throw away scripts to set some attributes in a collection of ufos. Feels nicer to have this as a reusable generic script.

m4rc1e:
Perhaps this could be generalised even further and handle .glyphs and binaries?

m4rc1e:
Merging since I need this. Will generalise it if I need to do similar stuff for .glyphs files.

PR #501, opened 2022-02-23
remove redundant scripts

m4rc1e:
**Gftools-build-font2ttf:**
Predates gftools builder. Also only works for fontforge sources.

**Gftools-build-vf**
Predates gftools builder.

**Gftools check-gf-github**
Use gftools-push-stats

**gftools-check-vf-avar**
Script was written before diffenator could handle vfs.

**gftools-dump-names**
Pretty much identical to gftools-check-names minus the csv stuff.

**Gftools-fix-dsig**
Drop since we don’t require dummy disks any more


m4rc1e:
**Gftools fix-vf-meta**
gftools builder does this automatically.

PR #502, opened 2022-02-23
Update gftools-gen-push-lists.py

RosaWagner:
I changed the label names so categorisation is more meaningful so reflected here accordingly


m4rc1e:
LGTM. I'll wait for the CI to pass then merge. Will cut a release tomorrow. Thank you.

PR #507, opened 2022-02-28
ufo-fix-instances: added

m4rc1e:
Fix designspace instances so they are gf spec complaint:
https://github.com/googlefonts/gf-docs/tree/main/Spec#fvar-instances


m4rc1e:
Once I've got a bit of time, I'll modularise these tools a bit more. Would be nice to chuck all the ufo/ds stuff into its own module.

PR #508, opened 2022-03-02
packager: get files from archives

m4rc1e:
Thought I'd open this before doing the full implementation since it may be controversial.

The aim of this PR is to allow users to fetch fonts/files from online archives such as Github releases. Imo font binaries should not be stored in repos. It is much nicer to create a tagged release which contains an archive which has the released font files e.g here's Roboto Serif's v1.007 release page:

Screenshot 2022-03-02 at 12 05 00

Users can then get the files from this release by using the `archive` field in the upstream yaml file e.g

```
branch: main
# Download url
archive: https://github.com/googlefonts/roboto-serif/releases/download/v1.007/RobotoSerifFonts-v1.007.zip
files:
# works in the same manner, just specify the dirs e.g
variable/RobotoSerif[wght].ttf: RobotoSerif[wght].ttf
...
```

If folks do want this feature, I'll improve it before asking for a review.

### DIFF
--- a/Lib/gftools/packager.py
+++ b/Lib/gftools/packager.py
@@ -41,11 +41,8 @@ from hashlib import sha1
 from fontTools.ttLib import TTFont # type: ignore
 from gflanguages import LoadLanguages
 from gftools.util import google_fonts as fonts
-<<<<<<< HEAD
 from gftools.github import GitHubClient
-=======
 from gftools.utils import download_file
->>>>>>> aca2234 (packager: get files from archives)
 
 # ignore type because mypy error: Module 'google.protobuf' has no
 # attribute 'text_format'
@@ -864,7 +861,8 @@ def _copy_upstream_files_from_dir(source_dir: str, files: dict,
 
 def _create_or_update_metadata_pb(upstream_conf: YAML,
                                   tmp_package_family_dir:str,
-                                  upstream_commit_sha:str) -> None:
+                                  upstream_commit_sha:str=None,
+                                  upstream_archive_url:str=None) -> None:
   metadata_file_name = os.path.join(tmp_package_family_dir, 'METADATA.pb')
   try:
     subprocess.run(['gftools', 'add-font', tmp_package_family_dir]
@@ -890,7 +888,10 @@ def _create_or_update_metadata_pb(upstream_conf: YAML,
   # metadata.date_added # is handled well
 
   metadata.source.repository_url = upstream_conf['repository_url']
-  metadata.source.commit = upstream_commit_sha
+  if upstream_commit_sha:
+    metadata.source.commit = upstream_commit_sha
+  if upstream_archive_url:
+    metadata.source.archive_url = upstream_archive_url
 
   language_comments = fonts.LanguageComments(LoadLanguages())
   fonts.WriteProto(metadata, metadata_file_name, comments=language_comments)
@@ -933,7 +934,7 @@ def _create_package_content(package_target_dir: str, repos_dir: str,
       print('DONE downloading and unzipping!')
       skipped = _copy_upstream_files_from_dir(tmp, upstream_conf['files'],
                         write_file_to_package, no_allowlist=no_allowlist)
-    upstream_commit_sha = upstream_conf['archive']
+      upstream_archive_url = upstream_conf['archive']
   else:
     local_repo_path_marker = 'local://'
     if upstream_conf['repository_url'].startswith(local_repo_path_marker):
@@ -1016,7 +1017,7 @@ def _create_package_content(package_target_dir: str, repos_dir: str,
 
   # create/update METADATA.pb
   _create_or_update_metadata_pb(upstream_conf, package_family_dir,
-                                upstream_commit_sha)
+                                upstream_commit_sha, upstream_archive_url)
 
   # create/update upstream.yaml
   # Remove keys that are also in METADATA.pb googlefonts/gftools#233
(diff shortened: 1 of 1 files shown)


simoncozens:
I'm planning on moving Noto binaries to being delivered via GitHub Releases rather than a repo full of binary files, so yes, I need this.

graphicore:
I believe I decided against adding this feature when I wrote the packager. We all agree, I believe, that having binaries in the repository is not ideal, but, having them as an attachment of a GitHub release is IMO even worse. One reason is that the releases and their attachments are a specific GitHub feature and hence a vendor lock-in. They are even more detached from the repo-history than binaries in git: an attachment doesn't relate at all to the git commit hash. In an ideal world we would build the fonts directly from the sources and record all dependencies required for a reproducible build, this, in my opinion establishes a pattern that is even worse than the status quo of having binaries in the repo, it is a step in the wrong direction.

anthrotype:
GitHub releases are associated with a tagged commit so they can be traced back to the source state at the given tag. It's true that they are a Github-specific feature, even though an handy one

simoncozens:
Note that the suggestion is to take the release from an arbitrary `archive` URL; in our cases this will be a GitHub release URL, but there is no GitHub lock-in here.

graphicore:
OK, no lock-in for arbitrary URLs, but then even more decoupling between sources and binaries.

graphicore:
> GitHub releases are associated with a tagged commit so they can be traced back to the source state at the given tag. It's true that they are a Github-specific feature, even though an handy one

We should use the tag directly then, and store the commit-hash of that, because tags can also change their target. GitHub release attachments can be anything, there's no relation to the source required at all. Sure they are handy, but I think this is not a good case, especially when we should move closer to building from the source not further away.

m4rc1e:
> GitHub release attachments can be anything, there's no relation to the source required at all.

Same argument could be made for the fonts which are stored in the /fonts dir. There's nothing stopping a designer from running the build chain then hand tweaking. This has happened in the past.

m4rc1e:
I reckon the majority of people who will be using this feature will be users of the Unified Font Repository. This repo uses a github action to build and automatically release the fonts. The font artifacts are attached automatically to the release without any human intervention.

graphicore:
> Same argument could be made for the fonts which are stored in the /fonts dir. There's nothing stopping a designer from running the build chain then hand tweaking. This has happened in the past.

True. There's still a timely hint having it fixed in history (Tags can change).  Ultimately reproducible building from source is the one real thing.

simoncozens:
> Ultimately reproducible building from source is the one real thing.

It's the ideal. If we were purist, we would be onboarding source repos and building them. I think we've already accepted the compromise of onboarding binaries. Whether we get the binaries from a repo or from a URL does not impact on the reproducible building question.

graphicore:
> [...] I think we've already accepted the compromise of onboarding binaries. Whether we get the binaries from a repo or from a URL does not impact on the reproducible building question.

Agreed. What I'm arguing is that allowing arbitrary URLs to zip archives is even worse than what we're doing now. In the case @m4rc1e made, of automated builds from UFR, it doesn't look so bad though. But that's a lot of context that the proposal doesn't cover.

graphicore:
> I'd be happy to restrict this so it only works for github releases archives if that makes you feel more comfortable?

I'm already comfortable with leaving this discussion as a trace.

m4rc1e:
Ok cool. Let me tidy up the implementation.

m4rc1e:
I've tried to add this feature as minimally as possible. 

The url is also added to the METADATA.pb `source.commit` field. Perhaps it would be better to request the METADATA protobuf gets its own archive field since it's not really a commit?

```
+source {
+  repository_url: "https://github.com/googlefonts/roboto-serif"
+  commit: "https://github.com/googlefonts/roboto-serif/releases/download/v1.007/RobotoSerifFonts-v1.007.zip"
+}
```

m4rc1e:
@graphicore I guess we could take a look at this.

graphicore:
Yeah, that's an interesting one. For a GitHub release, one could get the commit hash the tag was pointing to at that moment. I can think of a release based workflow where one would always release from a `latest` tag, but I'm not sure how well that would go with GitHub releases. It's possible to add new fields to a protobuf type without breaking backward compatibility (at least in the format, the consumer may still struggle) so the question is maybe which new fields should be in `source` (a `type` one ...). What purpose does `source` serve? Maybe answering that will help us to decide what fields are required.

m4rc1e:
I've decided to go with adding a new field, #517. If gf eng are ok with this, I'll update this pr.

simoncozens:
What are the expected files / structure of the release file? Does the packager look at all the files in the root, or fonts in fonts/ttf/ or ...? Does it look for any metadata files in the release zip? etc.

m4rc1e:
The packager produces upstream.yaml metadata files. The user has to manually input this information. The file includes `files` field which is a mapping that maps the upstream file path to the google/fonts/ofl dir paths. Here's Alegreya's https://github.com/google/fonts/blob/main/ofl/alegreya/upstream.yaml#L2-L5

We won't do anything differently for archives.

simoncozens:
What's the status of this?

m4rc1e:
I've just merged #517 so we should be good. Lemme just take one last look at this and rebase. Will be around half an hour or so.

simoncozens:
Thanks - no hurry, but I will be hoping to build stuff on top of it sometime soon.

PR #514, opened 2022-03-08
Move Github client from packager to own module

simoncozens:
The packager module is very big, and contains lots of code, not all of which is related to packaging; in addition, the GitHub-y bits of it are also used by gftools-qa. I'm hoping to use them for something else too, but they need disentangling first.

I've tested that gftools qa still works after this, but not the packager bits.

### DIFF
--- a/Lib/gftools/github.py
+++ b/Lib/gftools/github.py
@@ -1,0 +1,100 @@
+import os
+import pprint
+import requests
+import typing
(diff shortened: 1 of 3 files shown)


RosaWagner:
All the command I usually use are working fine

RosaWagner (approved):

PR #519, opened 2022-03-16
Merge two UFOs

simoncozens:
This is (hopefully) a generic tool to merge glyphs from one UFO into another. It takes care of some tricky cases:

* Added glyphs having components.
* Added glyphs get added to relevant lists/dicts in the font's lib.
* Kerning between two added glyphs.
* Kerning between added glyphs and existing glyphs.
* When added glyphs are produced by ligatures/substitutions, those layout rules get copied into the output UFO.
* Optionally, when added glyphs produce *other* glyphs by ligature/substitution, those glyphs together with their layout rules get added to the output UFO.


simoncozens:
@m4rc1e Ready for review, I think.

PR #516, opened 2022-03-09
refactoring to use external `axisregistry` module

felipesanches:
https://github.com/googlefonts/axisregistry
(issue #515)

Issue #515, opened 2022-03-09
Use external `axisregistry` module


felipesanches:
Similar to #510, now we have a python module for handling data from the Google Fonts Axis Registry.

https://pypi.org/project/axisregistry/

m4rc1e:
However, Nate clearly knows it exists ;-)

<img width="925" alt="Screenshot 2022-03-10 at 09 56 17" src="https://user-images.githubusercontent.com/7525512/157637428-f83bd219-883f-47a3-8a79-d72705c60c6d.png">

Guess it may just be a case of redirecting people instead.

PR #520, opened 2022-03-16
Fixing missing axisregistry items

aaronbell:
It appears that when the axisreg code was split out (see #515 #516 ), a couple of instances of axisreg were missed, causing gftools packager to now fail.

This PR corrects the two that I found.


felipesanches:
thanks, @aaronbell :-)

m4rc1e:
Thanks Aaron!

m4rc1e (approved):
LGTM

m4rc1e:
I see the `add-font` check doesn't actually run in `test_usage.py`. I'll fix this.

PR #521, opened 2022-03-18
[Builder] Make the schema inheritable

simoncozens:
I'm extending the builder for the Noto project; not an upstreamable extension - the idea is to fetch other Noto sources and merging subsets from them into the UFO while building. To do this, I need to extend the YAML schema to allow me to define the subsets in the config file. Currently I can't do that without major surgery. With this two-liner, the schema becomes inheritable.


m4rc1e:
Thanks!

PR #526, opened 2022-03-29
Add drop hints script from Noto

simoncozens:


bin/gftools-drop-hints.py
```diff
@@ -0,0 +1,64 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+"""Drop hints from a font."""
+
+__author__ = "roozbeh@google.com (Roozbeh Pournader)"
```
m4rc1e, line 20: He left a while ago so I don't think that email works anymore.

bin/gftools-drop-hints.py
```diff
@@ -0,0 +1,64 @@
...
+    """Drops the listed tables from a font."""
+    for table in tables:
+        if table in font:
+            del font[table]
+
+
+def main():
+    """Drop the hints from the first file specified and save as second."""
+    args = parser.parse_args()
+    font = ttLib.TTFont(args.input)
```
m4rc1e, line 52: `ttLib.TTfont` --> `TTFont`

m4rc1e:
Lemme change these since they're small.

bin/gftools-drop-hints.py
```diff
@@ -0,0 +1,61 @@
...
+parser.add_argument("output", help="file to save font", nargs="?")
+
+
+def drop_hints_from_glyphs(font):
+    """Drops the hints from a font's glyphs."""
+    glyf_table = font["glyf"]
+    for glyph_index in range(len(glyf_table.glyphOrder)):
+        glyph_name = glyf_table.glyphOrder[glyph_index]
+        glyph = glyf_table[glyph_name]
+        if glyph.numberOfContours > 0:
```
anthrotype, line 34: This is incomplete for it only drops instructions for simple glyphs, but not also for components in composite glyphs (where numberOfContours is -1)

Have a look at fonttools subsetter

PR #527, opened 2022-03-29
Add check-vertical-extents from nototools

simoncozens:
- Add "test-vertical-extents" (now called "check...") from nototools
- Add vharfbuzz dependency


m4rc1e (commented):
Thanks.

I've just tested this using the following cmd:

`gftools check-vertical-extents ~/Type/fonts/ofl/notosans/NotoSans-Regular.ttf`

It's taking a considerable amount of time. Is this normal? It may be nice to print a message mentioning it may take a long time.

simoncozens:
You need to feed it some text.

PR #528, opened 2022-03-30
Add some scripts to be used in GH Actions

simoncozens:
The idea here is that there are certain (complex) steps we need to perform as part of GitHub actions in the googlefonts-project-template repo. If those steps lived in the googlefonts-project-template repo itself, we would lose control over the ability to maintain them, as they would be forked into all of the downstream repos. But if the code to perform those steps is part of gftools, then we can roll out updates just by doing another gftools release. The GitHub actions don't use the venv (because that's used for font building and is frozen, so that won't get updates either); instead they explicitly grab the latest version of gftools and run the actions outside the venv. So they're always up to date as of when the Action is run.

The scripts I've added are:

* `checkgooglefonts`: Very simple - reads the `config.yaml` and sets a GH Actions workflow context if `config["googleFonts"]` is true (i.e. if this font is intended to be onboarded to GF.)
* `checkversionbump`: This one is not so simple. It checks whether the version of the font in the font source files is greater than the last tagged release; i.e. if the font has been bumped since the last release was created, and therefore if a new release is needed. It sets GH Actions workflow contexts to indicate if a new release is needed and if so what version number / tag it should be.
* `notobuilder`: This is a subclass of gftools builder which I'm planning to use as part of the Noto build process. It's not plugged into anything yet; that's the next piece of the puzzle but I wanted to get the onboarding stuff working first.
* `qa2issue`: This takes a fontbakery report *and a version number*. It then looks on Github for an issue related to the version; if there is not one, it creates one, and if there is one, it adds the report as a comment. This ensures that each font version gets one single QA issue, so that the user does not get spammed with new issues for each commit.
* `updateupstream`: This updates an upstream.yaml from a GitHub release, generating one if there isn't one in the repo already.

Note that this introduces a new schema key to `config.yaml`

### DIFF
--- a/Lib/gftools/actions/__init__.py
+++ b/Lib/gftools/actions/__init__.py
@@ -1,0 +1,7 @@
+"""Scripts used for font repository automation by GitHub Actions
+
+This contains utility scripts used within GitHub Actions. Putting them
+here allows us to ensure that this code doesn't live in individual
+font repositories, and that the GH Actions can always access the most
+up-to-date versions of the scripts by installing gftools as part of the
+action."""
--- a/Lib/gftools/actions/checkgooglefonts.py
+++ b/Lib/gftools/actions/checkgooglefonts.py
@@ -1,0 +1,15 @@
+"""Test whether the font should be PRed to GF, inform GitHub if so.
+"""
+import yaml
+import os
+from sys import exit
+
+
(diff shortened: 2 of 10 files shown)


m4rc1e:
Thanks!

PR #529, opened 2022-03-30
Add instruction to upstream template yaml

RosaWagner:
Emma is trying to use the feature, but couldn't find instruction, is this correct @m4rc1e ??


m4rc1e (commented):
Thank you. You're hitting these prs out the park these days.

Yep, she's correct. To get the zip url, I usually right click on it

Screenshot 2022-03-30 at 16 23 14

I'll do a quick test and merge if we're good.

PR #532, opened 2022-04-04
add support for axis illustration files in push.py

zhaoxiong0211:


Lib/gftools/push.py
```diff
@@ -98,11 +98,13 @@ def push_report(fp):
 def missing_paths(fp):
     paths = [fp.parent / p.path for p in parse_server_file(fp)]
     axis_files = [p for p in paths if p.name.endswith("textproto")]
+    axis_illustration_files = [p for p in paths if p.name.endswith("svg")]
```
m4rc1e, line 101: `endswith` can take a tuple, https://docs.python.org/3/library/stdtypes.html#str.endswith. This is cleaner:

```
axis_files = [p for p in paths if p.name.endswith(("textproto", "svg))]
```

zhaoxiong0211: thank you! resolved

m4rc1e:
Apologies but I'm not aware of what you're working on. Are we adding illustrations for each axis? will they be visible on fonts.google.com?

zhaoxiong0211:
> Apologies but I'm not aware of what you're working on. Are we adding illustrations for each axis? will they be visible on fonts.google.com?

sorry i should have linked this PR. https://github.com/google/fonts/pull/4470/files
We are aiming to add illustration to each axis and it shall be shown to users together with the description

m4rc1e:
Ok cool. I'm happy with this pr.

We currently have two axis registries which is not ideal. One in google/fonts, the other over at https://github.com/googlefonts/axisregistry. @felipesanches recently made the latter. @zhaoxiong0211 would it be possible to also add your google/fonts pr to this repo? Ideally, we should drop the axis registry in google/fonts and just use the repo in the future.

zhaoxiong0211:
PR in googlefonts/axisregistry
https://github.com/googlefonts/axisregistry/pull/29

nathan-williams (approved):

PR #534, opened 2022-04-06
Use textprotos for sample text if none found in UDHR

### LINKED ISSUE #533: Use googlefonts/lang for sample text in gen-html
I'm using a googlefonts-project-template fork for Noto, and want to build up some proof sheets using `make proof`. For many scripts this won't work because the scripts aren't in `uhdr_all.txt`. However, we do have sample texts for each language in the textprotos.

I'm be interested in making a PR which uses the languages repository to grab a sample text for a font if nothing could not be found in the uhdr. How does that sound?

simoncozens:
Fixes #533

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -22,6 +22,7 @@ import shutil
 import unicodedata
 from unidecode import unidecode
 from collections import namedtuple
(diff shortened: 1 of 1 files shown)


m4rc1e (approved):
LGTM. Thanks.

PR #537, opened 2022-04-08
builder: run filters in ufo lib.plist

m4rc1e:
Depends on https://github.com/googlefonts/fontmake/pull/874, https://github.com/googlefonts/ufo2ft/pull/604

Fixes https://github.com/googlefonts/fontmake/issues/872

cc @yanone @anthrotype

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -313,11 +313,11 @@ class GFBuilder:
             tmpdir = tempfile.TemporaryDirectory()
             args["output_dir"] = tmpdir.name
 
+        filters = args.get("filters", [])
         if (
             self.config["flattenComponents"] or
             self.config["decomposeTransformedComponents"]
         ):
-            filters = args.get("filters", [])
             if self.config["flattenComponents"]:
                 filters.append(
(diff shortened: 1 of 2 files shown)


anthrotype:
@yanone it's deliberately not a string, it's the `Ellipsis` object => https://docs.python.org/3.9/library/constants.html#Ellipsis

it's just a placeholder to say something was omitted and will be filled in later in the pipeline.

simoncozens:
I don't think it is in this case. I think it's meant to be the Python equivalent of the command line `--filter='...'` you just added. So it should be a string.

simoncozens:
Ah, OK. Well, Marc's code is right then.

anthrotype:
FYI https://github.com/googlefonts/fontmake/releases/tag/v3.3.0

m4rc1e:
Thanks Cosimo, I'm rerunning the CI now and it's looking good!

anthrotype:
should you not also bump the minimum required fontmake in setup.py? 
https://github.com/googlefonts/gftools/blob/12a0e7f81ead117e2ad89819f0451b5db1cc4e93/setup.py#L95

m4rc1e:
Good catch.

PR #539, opened 2022-04-14
Move QA class to its own library for code sharing

simoncozens:
I'd like to make a gftools-qa-like tool for Noto, which downloads the latest release of a family from GitHub releases and compares it as well as running fontbakery tests. I've since realised that I can do most of this by using `-ghb` on the "distribution" repo if I restructure it a bit, but it's still a good idea to separate front-end/back-end and make things easily inheritable if they can be.


bin/gftools-qa.py
```diff
@@ -35,244 +33,12 @@
     mkdir,
 )
 from gftools.html import HtmlProof, HtmlDiff
```
m4rc1e, line 35: This import needs moving. Can do this if you'd like?

simoncozens: Oops. Yes, thanks!

m4rc1e (approved):
Very nice!

I like that you've added parameters for running fontbakery

I'll merge and cut a new release once the ci has finished.

PR #543, opened 2022-04-18
Allow gftools-qa to download fonts from a URL

simoncozens:
Now we are using GitHub release more, it becomes useful to QA a font against the previous release. e.g.

```
$ bin/gftools-qa.py -arb https://github.com/notofonts/limbu/releases/download/NotoSansLimbu-v2.002/limbu-2.02.zip -f fonts/ttf/Limbu-Regular.ttf --diffenator
```

This implements the `-arb` argument to gftools-qa.

### DIFF
--- a/Lib/gftools/utils.py
+++ b/Lib/gftools/utils.py
@@ -197,6 +197,12 @@ def download_files_in_github_dir(
     return results
 
 
+def download_files_from_archive(url, dst):
+    zip_io = download_file(url)
+    with ZipFile(zip_io) as zip_file:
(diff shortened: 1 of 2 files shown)


m4rc1e (commented):
Nice.

What about adding the arg `-ar` which will allow us to get fonts from an archive which are not the before variants?

simoncozens:
Good idea.

m4rc1e (approved):
LGTM.

In the future we may want to refactor `fonts_from_zip` so we can specify which fonts we want. Atm it takes all ttfs.

PR #545, opened 2022-04-20
builder: ensure fontmake args are reset when generating fonts from each source file

m4rc1e:
If a config file contains 2 or more sources, the second sources onwards will inherit the previous source's args. We want fresh args for each source. 

Hence why the ci complains in https://github.com/eliheuer/hasubi-mono/runs/6088671254?check_suite_focus=true

Fixes https://github.com/googlefonts/googlefonts-project-template/issues/72

Apologies for crappy explanation. Haven't slept well and I'm having a new floor laid. It's louder than a Dutch Gabber rave.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -455,10 +455,7 @@ class GFBuilder:
                 "output_dir": directory,
                 "optimize_cff": CFFOptimization.SUBROUTINIZE,
             }
-            if source.endswith("ufo"):
-                if "interpolate" in args:
-                    del args["interpolate"]
-            else:
+            if not source.endswith("ufo"):
                 args["interpolate"] = True
             self.logger.info("Creating static fonts from %s" % source)
             for fontfile in self.run_fontmake(source, args):
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -279,9 +279,9 @@ class GFBuilder:
 
     def build_variable(self):
         self.mkdir(self.config["vfDir"], clean=True)
-        args = {"output": ["variable"], "family_name": self.config["familyName"]}
         ttFonts = []
         for source in self.config["sources"]:
+            args = {"output": ["variable"], "family_name": self.config["familyName"]}
             if not source.endswith(".designspace") and not source.endswith("glyphs"):
                 continue
             self.logger.info("Creating variable fonts from %s" % source)
@@ -449,12 +449,12 @@ class GFBuilder:
 
     def build_a_static_format(self, format, directory, postprocessor):
         self.mkdir(directory, clean=True)
-        args = {
-            "output": [format],
-            "output_dir": directory,
-            "optimize_cff": CFFOptimization.SUBROUTINIZE,
-        }
         for source in self.config["sources"]:
+            args = {
+                "output": [format],
+                "output_dir": directory,
+                "optimize_cff": CFFOptimization.SUBROUTINIZE,
+            }
             if source.endswith("ufo"):
                 if "interpolate" in args:
                     del args["interpolate"]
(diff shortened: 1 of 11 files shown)


m4rc1e:
@simoncozens Lemme add a test for the builder since we don't have any.

m4rc1e:
Just checked the current gftools on pypi on https://github.com/googlefonts/Libre-Bodoni and it fails ot build. This PR works so I'll add this family as a testcase.

Lib/gftools/builder/__init__.py
```diff
@@ -449,12 +449,12 @@ def instantiate_static_fonts(self, directory, postprocessor):
...
-            "optimize_cff": CFFOptimization.SUBROUTINIZE,
-        }
         for source in self.config["sources"]:
+            args = {
+                "output": [format],
+                "output_dir": directory,
+                "optimize_cff": CFFOptimization.SUBROUTINIZE,
+            }
             if source.endswith("ufo"):
                 if "interpolate" in args:
```
simoncozens, line 459: This can probably go now, since there's no way for `interpolate` to get into `args`.

simoncozens (changes requested):
Yeah, this is fine. The current codepath is a bit confusing because it's not entirely clear what is modifying `args`. It took me a while to realise it was `run_fontmake`, which I didn't expect to have any side-effects. I wonder if *as well* we should use a separate args dictionary inside `run_fontmake` to make it side-effect-free.

(But that can be a clean-up for another time. This PR fixes the problem and should be merged.)

Lib/gftools/builder/__init__.py
m4rc1e: The `interpolate` argument in fontmake is a strange one. If I remove this arg, it won't generate glyphsapp instances for glyphsapp sources.

from --help:
```
 -i [INSTANCE_NAME], --interpolate [INSTANCE_NAME]
                        Interpolate masters and generate all the instances
                        defined. To only interpolate a specific instance (or
                        instances) that match a given "name" attribute, you
                        can pass as argument the full instance name or a
                        regular expression. E.g.: -i "Noto Sans Bold"; or -i
                        ".* UI Condensed". (for Glyphs or MutatorMath sources
                        only).

```

I'm going to change the code from:

```PYTHON
            if source.endswith("ufo"):
                if "interpolate" in args:
                    del args["interpolate"]
            else:
                 args["interpolate"] = True
```

to: 

```PYTHON
           if source.endswith("glyphs"):
               args["interpolate"] = True
```

This way we still get glyphsapp instances

Lib/gftools/builder/__init__.py
```diff
@@ -455,10 +455,7 @@ def build_a_static_format(self, format, directory, postprocessor):
                 "output_dir": directory,
                 "optimize_cff": CFFOptimization.SUBROUTINIZE,
             }
-            if source.endswith("ufo"):
-                if "interpolate" in args:
-                    del args["interpolate"]
-            else:
+            if source.endswith("glyphs"):
                 args["interpolate"] = True
```
simoncozens, line 459: Do we need to add the `-i` flag on .designspace sources? (I don't know.)

anthrotype: yes, if you want the instances, otherwise you get only the masters.

m4rc1e: I think the previous logic is ok. If it's a ufo, it'll delete the interpolate key if it exists, otherwise it'll set it for a ds or glyphs file. You ok for me to rollback my last commit and leave this alone?

m4rc1e: Right, I'm being a clown here. Just realised we're constructing the args a fresh in the loop just above. Ofc I'll just drop it for ufos.

m4rc1e: this should be good.

simoncozens (approved):
That's the one.

PR #547, opened 2022-04-23
ufomerge.py: Fix handling of missing font path

joshuakraemer:
`ufo2._path` can be `None`, which makes `Path(ufo2._path)` fail.


m4rc1e (approved):
LGTM. Apologies for the delay.

m4rc1e:
Will merge once CI has finished.

PR #548, opened 2022-04-26
add new QA HTML Test tool for comparing a font family.

ollimeier:
Hi there,

this PullRequest is related to a new test tool to compare the character set of a given font family. Please see the short demo video:

https://user-images.githubusercontent.com/21055547/165339403-7f28806f-ce55-4382-822d-6d4674892a5e.mp4

### DIFF
--- a/Lib/gftools/html.py
+++ b/Lib/gftools/html.py
@@ -22,6 +22,7 @@ from gftools.utils import (
     font_is_italic,
     partition,
     get_encoded_glyphs,
+    get_encoded_glyphs_from_fonts,
 )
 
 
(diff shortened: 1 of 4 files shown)


moontypespace:
@m4rc1e What do you think about this idea? Would be great to get your thoughts on that, because you have been responsible for the HTML test tools mainly, right?  
I use the kind of comparison almost everyday and thought it would fit perfectly into the google html test tools.
Would be great to extend it with unencoded glyphs and meta-data information as well, but for the first version I assume it's ok like it is for getting your feedback.
Thanks, Olli

m4rc1e:
Thanks Olli, this looks great!

I'll take a proper look at this next week.

moontypespace:
> I've been toying with the idea of animating through a VF's axes using css. I feel this may be a better approach than relying on just fvar instances since we have many families with multiple axes arriving.

Excuse me @m4rc1e , but I am not sure if I understand you right. Can you please explain it in different words?

moontypespace:
> Something like https://codepen.io/michellebarker/pen/bPEWGK, but for every character individually.

Absolutely :) I have done something similar already in a different context. I was also thinking about having some interactive functions included to de/activate opentype feature, increase font size, etc. This pull request is just the first draft and it would be great if we could extend it in future.
Have you seen this page: https://www.fontspecimen.com/ (I have done it during my time at Monotype). A lot of the ideas could be useful in the family.html as well.

moontypespace:
@m4rc1e Any idea, why the checks have restarted today and why they fail? A couple days ago, when I first made a pull request all checks passed.

moontypespace:
Hey @m4rc1e , is there anything I can help you with to finish this first version?

PR #555, opened 2022-05-12
Maintain Noto metadata

simoncozens:
Two fields in a font's METADATA.pb are used by Noto fonts. These need to be preserved when upgrading a font with add-font.


m4rc1e (approved):
LGTM

PR #567, opened 2022-05-27
Update gftools-gen-push-lists.py

RosaWagner:
Updating labels


RosaWagner:
The labels already exists

PR #582, opened 2022-06-30
Add gftools-autohint.py

simoncozens:
I realised we needed a command line wrapper around gftools.autohint (i.e. ttfautohint-py with our script-detection logic added) for the ninja builder, because everything the ninja builder calls has to be a command line utility.

But I also realised that it's useful for us to provide a command-line wrapper around ttfautohint-py *anyway*. The whole point of ttfautohint-py is that it gives you an easy way to install the latest version of ttfautohint, not just the one in your distribution (which can be quite old), and the whole point of gftools is that you can install it and get the latest version of everything you need to build fonts. But installing ttfautohint-py doesn't give you an easy way to run the latest version of ttfautohint.

I also needed something to get the right script for auto hinting full Noto fonts; these contain a Latin set but may be quite small. With our current script detection logic a large Latin set might overwhelm a small other set of glyphs, which are actually the ones you want to target for autohinting. The `discount_latin` argument ignores Latin glyphs when determining which is the primary script of a font .


m4rc1e:
Apologies for not reviewing this. I'll cut a release and look at this next week.

simoncozens:
No problem. I suggest we split the primary script detection out to another module, and also use the GSUB closure trick I posted elsewhere, so we can use this to fill in primary script metadata on new fonts (or backfill the old catalogue).

simoncozens:
I'm merging this because we have already added the primary script detection code, and we have builder.autohint already, so this is now just a CLI wrapper.

PR #583, opened 2022-06-30
stat: improvements

m4rc1e:
- use axisregistry.build_stat
- Only set the Variations PostScript Prefix name if one doesn't exist in sources. I've also moved this to fix_font so it isn't part of generating a STAT table anymore (as it shouldn't be).

@RosaWagner mind testing this?


RosaWagner:
It works as expected :)

PR #590, opened 2022-07-27
[packager] Add article/ARTICLE.en_us.html to allowlist

simoncozens:
This adds articles to the packager's allowlist. I need them for Noto and I had a few options here:

* Just keep this to my Noto gftools branch: possible but I'm trying to reduce divergence.
* Use `--no-allowlist`: but sanity checking is a good thing!
* Allow it for everyone: least bad option, probably won't be used by anyone other than Noto, but also it allows for the future possibility of having articles about library fonts.


m4rc1e (approved):
LGTM. If you've already used this on a Noto pr and it worked, feel free to merge.

PR #600, opened 2022-08-15
gen-push-lists: small fixes

m4rc1e:
- Do not include include noto article dirs
- fix lang directory path
- order push lists

cc @vv-monsalve

### DIFF
--- a/bin/gftools-gen-push-lists.py
+++ b/bin/gftools-gen-push-lists.py
@@ -74,7 +74,9 @@ def main():
 
             labels = set(l.name for l in content.labels)
             pr = content.as_pull_request()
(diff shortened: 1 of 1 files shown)


m4rc1e:
CI has failed due to an issue with fonttools. I'll merge this and raise/fix the fonttools issue.

vv-monsalve:
I've tested this today. 

> - Do not include include noto article dirs

Project + article path were listed, but this time the PR actually reinstates the article, so unsure if that should be the expected behaviour

> - fix lang directory path

Lang PRs are listed now repeated (good + bad path) 

> - order push lists

Working nice!

vv-monsalve:
After using it a couple of times now lang paths are working ok, but the article still is repeated.

PR #601, opened 2022-08-18
Make UFO merge more deterministic

Hoolean:
Currently, the output of the merge process is non-deterministic: if the tool is run twice with the same input UFOs, the outputs will differ. This is undesirable, as it produces lengthy and unnecessary UFO diffs, and makes the overall process the merge is part of harder to test.

## Cause

This originates when values are represented as sets: the existing order of properties is discarded, and the Python implementation is at liberty to choose whichever order it deems most convenient from run to run.

## Fix

While using sets is desirable – they are the most appropriate data structure – we wish to limit the number of orders they can take on when being converted back into lists. This commit converts the sets back into lists through `sorted()` instead of `list()`, which makes the order at runtime deterministic.

### Caveats

These changes stop short of normalisation: while often desirable, it is outside the scope of this tool, and is best left to *ufonormalizer* and friends. In contrast, we cannot do these changes upstream in *ufonormalizer*, as it preserves the order of certain fields that legacy font editors rely on (but which UFO merge already implicitly discards).

In addition, there are further uses of sets left unsorted; it is unclear whether their order at runtime affects output.

### Testing

This PR appears to function as expected following limited local testing but would benefit from automated test coverage or exposure to further UFOs : )

### DIFF
--- a/Lib/gftools/ufomerge.py
+++ b/Lib/gftools/ufomerge.py
@@ -65,10 +65,12 @@ def merge_ufos(
                 flat_outputs = [
                     item for sublist in rule.replacement for item in sublist
                 ]
-                true_inputs = [list(set(r) & newglyphset) for r in rule.input]
-                rule.precontext = [list(set(r) & newglyphset) for r in rule.precontext]
+                true_inputs = [sorted(set(r) & newglyphset) for r in rule.input]
+                rule.precontext = [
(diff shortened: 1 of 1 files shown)


simoncozens:
I think this is now fixed in (upstream) ufomerge.

PR #607, opened 2022-09-01
Try harder to get more sample text

simoncozens:
Currently we use UDHR to get some sample text, and then look in gflanguages' sample text collection if we got *nothing*. But we hardly ever get nothing at all, because there may be punctuation etc. in the font. (Even spaces!)

So this is more aggressive, and uses the gflanguages sample texts to try and get more text if we don't have complete coverage of the font's encoded characters.


m4rc1e:
Love it.

PR #608, opened 2022-09-14
Clean up and fix Lang category

RosaWagner:
I tested it locally and it works as expected.

- All small fixes are going in Upgrade category
- Lang stuff are displaying in Sample texts category
- Axis registry only having axis registry paths
- Anything else goes in Other

cc @m4rc1e @vv-monsalve


m4rc1e:
Nice! thank you and welcome back!

PR #610, opened 2022-09-20
fix: wip add fix_colr_font

m4rc1e:
Improvements I should make:

-  I had difficulties adding an empty glyph so I'm expecting the font to have one. It's not as simple as just adding a new glyph to the glyf table and then updating the font's glyphOrder order. You also need to update other tables such as hmtx etc,

cc @anthrotype basics are down. May chuck in CFF since I see Yanone has worked around this by using an area pen.


requirements.txt
```diff
@@ -28,3 +28,4 @@ hyperglot
 fontFeatures>=1.6.2
 vharfbuzz
 bumpfontversion
+nanoemoji
```
anthrotype, line 31: best to require a relatively recent version (the latest is preferable), use `>=`

Lib/gftools/tests/test_fix.py
```diff
@@ -237,3 +242,22 @@ def test_fix_vertical_metrics_typo_metrics_enabled(static_fonts):
...
+    fixed = fix_colr_font(colr_font)
+    assert "SVG " not in fixed
+    gid1 = fixed.getGlyphOrder()[1]
+    assert fixed["glyf"][gid1].numberOfContours == 0
+
+
+def test_fix_colr_v1_font(colr_font):
+    # Fix a COLR v1 font. Only maximum_color should be run
+    from gftools.fix import fix_colr_font
+    colr_font["COLR"].version = 1
```
anthrotype, line 261: you're loading a COLRv0 font (CairoPlay) but then setting its version to 1.. that's not sufficient for a valid conversion from version 0 to 1; there are additional attribtues that you're ignoring, also the fonttools API for v0 and v1 is slightly different for historical reasons (we kept backward compat for the v0 one to not break existing tools). I would suggest that in your pytest fixture you actually load a real COLRv1 test font (a small subset is fine).

Lib/gftools/fix.py
```diff
@@ -605,28 +605,30 @@ def drop_mac_names(ttfont):
...
-        new_order.insert(1, empty_glyph)
-        reorder_glyphs(ttfont, new_order)
+
+    gid1 = glyph_order[1]
+    if glyf_table[gid1].numberOfContours == 0:
+        return
+    _ensure_ttfont_fully_decompiled(ttfont)
+    hmtx = ttfont["hmtx"]
+    blank_glyph = Glyph()
+    blank_name = "blank"
```
anthrotype, line 621: the name "blank" might already exist, you don't want to overwrite anything. Try to come up with a more unique name. ".null" if present is very likely to be empty, if no empty glyphs are found in the font, then naming the second glyph ".null" seems like a safer option than "blank"

Lib/gftools/fix.py
```diff
@@ -605,28 +605,30 @@ def drop_mac_names(ttfont):
...
+    glyf_table[blank_name] = blank_glyph
```
anthrotype, line 622: the problem with using `_g_l_y_f_table.__setitem__` method is that it automatically appends to the TTFont's  glyphOrder list without also invalidating the {glyphNames:glyphID} cache, which means TTFont.getGlyphID could return incorrect result. Instead you can set the new glyphs directly inside the glyf's `glyphs` dict and call TTFont.setGlyphOrder at the end with the new desired order, which automatically triggers a rebuild of glyphID cache.
We do the same in nanoemoji.glue_together module where we add the layer glyphs used by COLR table when we maximum_color converts from OT-SVG => COLR:
https://github.com/googlefonts/nanoemoji/blob/2ffdd23faad153c5cec763b6e210befc9d704cd1/src/nanoemoji/glue_together.py#L53-L72
Also see related fonttools issue: https://github.com/fonttools/fonttools/issues/2605

Lib/gftools/tests/test_fix.py
m4rc1e: Thanks. I'll use real fonts.

Lib/gftools/fix.py
```diff
@@ -605,28 +605,30 @@ def drop_mac_names(ttfont):
...
+    ttfont["glyf"] = glyf_table
+    hmtx.metrics[blank_name] = (0, 0)
+    if "HVAR" in ttfont:
+         hvar = ttfont["HVAR"]
+         hvar.table.AdvWidthMap.mapping[blank_name] = 0
```
anthrotype, line 627: A few things:

1) `AdvWidthMap` is optional in HVAR, it may not be there at all (i.e. set to `None`). It happens to be in your test font, but it's not guiaranteed to be present. You should do like the subsetter does, and check `if hvar.table.AdvWidthMap:` before proceeding to update it

2) HVAR can optionally also include two additional mappings, `LsbMap` and `RsbMap` for the sidebearings (most font don't, but the spec allows those too and you should handle them here).

2) by using `0` for the empty glyph map, you are assuming that the HVAR.VarStore.VarData[0].Item[0] contains all zeros, that is no variation. This happens to be with the HVARs that fontTools.varLib generates, but it's not guaranteed that the first item in the ItemVariationStore is a no-op row of zero deltas. So it'd be safer to map it to `0xFFFFFFFF` which in OT means "no variation". You can import the constant from `fontTools.ttLib.tables.otTables import NO_VARIATION_INDEX`

m4rc1e: Thanks!

Lib/gftools/fix.py
m4rc1e: Thanks. I will explore this now.

Lib/gftools/fix.py
```diff
@@ -609,24 +609,33 @@ def _add_blank_glyph_to_gid1(ttfont):
...
 
+    if len(glyph_order) < 2:
+        return
     gid1 = glyph_order[1]
     if glyf_table[gid1].numberOfContours == 0:
         return
     hmtx = ttfont["hmtx"]
     blank_glyph = Glyph()
-    blank_name = "blank"
+    blank_name = ".null" if ".null" not in glyph_order else "emptyglyph"
```
anthrotype, line 624: you may want to `assert "emptyglyph" is not in glyph_order` maybe just to be on the safe side.

anthrotype (approved):
LGTM, with a few nits:

- use subprocess.run(check=True) to make sure it stops if something goes bad
- test that _add_empty_glyph_to_gid1 method does get called by fix_colr_v0_gid by making a subset test font without empty glyphs
- maybe factor out logic that adds SVG table to COLR font to its own distinct function

m4rc1e:
Thanks Cosimo. You've been great

PR #612, opened 2022-09-21
Added more entries in add-axis script

RosaWagner:
I tried to do something but it doesn't work ^^
I open the PR so we can debug tomorrow @m4rc1e if you have time.

```
  File "/Users/rosalie/Google/env2/bin/gftools-add-axis.py", line 162, in <module>
    main(**args.__dict__)
  File "/Users/rosalie/Google/env2/bin/gftools-add-axis.py", line 106, in main
    axis_proto.fallback_only = false
NameError: name 'false' is not defined
```

I see the problem but don't know how to solve yet.

### DIFF
--- a/bin/gftools-add-axis.py
+++ b/bin/gftools-add-axis.py
@@ -104,8 +104,6 @@ def main(font: str):
     axis_proto.precision = 1  # ask user?
     # Wether or not fallback positions should be used (but not sure)
     axis_proto.fallback_only = False
-    # Illustration path
-    axis_proto.illustration_url = ''
     # Short descriptive paragraph
     axis_proto.description = (  # ask user?
         'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod'
@@ -131,6 +129,7 @@ def main(font: str):
     filename = f'{axis_proto.display_name.lower()}.textproto'
     with open(filename, 'x') as f:
         f.write(text_proto)
+    print(f"Please remember to link an illustration. The description should be reviewed by a Knowledge content editor.")
     print(f'DONE create {filename}!')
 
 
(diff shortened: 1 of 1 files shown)


bin/gftools-add-axis.py
```diff
@@ -102,6 +102,10 @@ def main(font: str):
     axis_proto.max_value = fvar_axis.maxValue
     # Input values for this axis must aligned to 10^precision
     axis_proto.precision = 1  # ask user?
+    # Some parameter I don't know what it does
+    axis_proto.fallback_only = false
```
vv-monsalve, line 106: Would it always be false? I understand not.

RosaWagner: I don’t know, I wanted to ask more details about it to you at the meeting, and the eng team if needed. I just see that all textprotos are having that parameter now with false as value so I thought it could be the default in this script?

bin/gftools-add-axis.py
```diff
@@ -102,6 +102,10 @@ def main(font: str):
...
+    # Illustration path
+    axis_proto.illustration_url = '' # leave empty?
```
m4rc1e, line 108: Wouldn't it be better if the tool could take an illustration as an arg?

RosaWagner: We probably do the illustration in a second step once this axis is already in sandbox

m4rc1e: Arg could be optional.

m4rc1e (commented):
Only concern is defaulting to False for `axis_proto.fallback_only`. I may accept this once we know what this field does. Otherwise it's fine.

bin/gftools-add-axis.py
vv-monsalve: Evan said: The fallback_only was added to communicate to the front end how to present the axis in the UI. If set to true, the UI would present it as a dropdown (or radio buttons) instead of a slider. Cursive is the only axis currently using it. 

Eventually, the slider option could be set as default in the API, and the field would only be added when in need to override it (similar to the default value). But until the API is updated, it seems to leave it predefined as "false" here would be ok.

PR #617, opened 2022-10-03
Expose more fontmake args in the config

### LINKED ISSUE #301: Add option to builder to pass flags to fontmake
As mentioned on call. This may be tricky because we don't use fontmake's command line parser, so we may have to parse the arguments ourselves.

### LINKED ISSUE #558: [Builder] Add "expand-features-to-instances" option to config.yaml
I'm currently using `gftools builder` to build from source UFOs where the `features.fea` file contains an `include(<filename>)` directive. The instance UFOs are unable to resolve this path and the build process fails with the error `The following feature file should be included but cannot be found`

Fontmake has an `--expand-features-to-instances` option - it would be good to specify this as an option in the `config.yaml` when building.

### LINKED ISSUE #581: [builder] Can the builder accept FontMake args?
## Context
I have a problem with features and building variable plus static fonts:
1. I want to write fea code in only one place, then reference that from multiple UFOs, using fea code like `include(./features/locl.fea);` (which would reference a file within a `features` directory at the same level as the UFO
2. The relative reference paths are different for variable builds and static builds. For variable builds, the references can be at the same level as the UFO, like `./features/locl.fea`. However, for static builds, FontMake starts by interpolating UFOs and puts them into the subdirectory `instances`, so relative paths then need to be like `../features/locl.fea` to go up one level, then into the `features` directory.

FontMake has a solution for this via the arg `--expand-features-to-instances`:
```
expand_features_to_instances: parses the master feature file, expands all
include()s and writes the resulting full feature file to all instance
UFOs. Use this if you share feature files among masters in external
files. Otherwise, the relative include paths can break as instances
may end up elsewhere. Only done on interpolation.
```

I am specifically trying to use the [Ninja Builder](https://github.com/googlefonts/gftools/pull/569) to see if it will build a large family more quickly than my current shell script. My script is using FontMake to build 108 static fonts, so things are taking 45 minutes or more. 😬 Making it faster would really help my workflow for debugging technical issues reported by FontBakery, etc.

## Questions

1. Is there a way to feed fontmake args to the gftools builder? 
    1. If not, is there another simple way to solve my issue? I guess I could use a script to expand the features into my UFOs in a pre-build step, if nothing else... Is there an existing script for this, or should I look at how FontMake does it and adapt that into my own script? I'm happy to get suggestions here, even if you don't know a full answer!

2. Can I give the builder args if I am pointing it to a designspace rather than a `config.yaml` file? For example, can I do something like: `gftools builder sources/family.designspace --outputdir fonts-gftools-builder`?
    1. If not, is there a semi-automated way to start a config.yaml file from a designspace?

madig:
Expose the reverse outlines, remove overlaps and expand features to instance arguments in the config.yml.

Expanding features is on by default because it's the only default that makes sense.

I chose to only expose these three, because GFBuilder calls fontmake's internals directly, bypassing the argument parser. I don't want to duplicate the logic of the parser.

Closes #301 
Closes #558 
Closes #581

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -99,6 +99,9 @@ required, all others have sensible defaults:
 * ``googleFonts``: Whether this font is destined for release on Google Fonts. Used by GitHub Actions. Defaults to ``false``.
 * ``category``: If this font is destined for release on Google Fonts, a list of the categories it should be catalogued under. Used by GitHub Actions. Must be set if ``googleFonts`` is set.
 * ``fvarInstanceAxisDflts``: Mapping to set every fvar instance's non-wght axis
+* ``expandFeaturesToInstances``: Resolve all includes in the sources' features, so that generated instances can be compiled without errors. Defaults to ``true``.
+* ``reverseOutlineDirection``: Reverse the outline direction when compiling TTFs (no effect for OTFs). Defaults to fontmake's default.
+* ``removeOutlineOverlaps``: Remove overlaps when compiling fonts. Defaults to fontmake's default.
 value e.g if a font has a wdth and wght axis, we can set the wdth to be 100 for
 every fvar instance. Defaults to ``None``
 """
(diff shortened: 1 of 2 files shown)


Lib/gftools/builder/__init__.py
```diff
@@ -334,6 +334,13 @@ def run_fontmake(self, source, args):
         # https://github.com/googlefonts/fontmake/issues/872
         args["filters"] = [...] + filters
 
+        if source.endswith((".glyphs", ".designspace")):
+            args["expand_features_to_instances"] = self.config.get(
+                "expandFeaturesToInstances", True
+            )
+        args["reverse_direction"] = self.config.get("reverseOutlineDirection", True)
+        args["remove_overlaps"] = self.config.get("removeOutlineOverlaps", True)
+
```
m4rc1e, line 343: I'm assuming that by not setting these fields, we'll get fontmake defaults? if we don't, we'll end up breaking existing projects.

madig: Yas, removing overlaps and reversing direction are the default in fontmake. Expanding the features is off by default, but projects are more likely to break that way and turning it on is the only sensible option I think.

belluzj: If the fields are not in self.config, should we maybe not pass anything to fontmake, and let fontmake fill-in the default values? That way the default values are stored in only one place, fontmake.

madig: Done.

m4rc1e:
Thanks. I'll cut a new release for you later.

PR #620, opened 2022-10-12
Font primary script utilities

simoncozens:
This adds two things:

* a `primary_script` function in gftools.utils which finds the primary script of a font, optionally ignoring Latin characters. It does this by categorising all encoded glyphs by their script and also applying the same script to any other glyphs produced by substitutions. It also rewrites the `autohint_script_tag` function to use that new function.
* a `gftools-set-primary-script` command line utility which walks a directory tree looking for METADATA.pb files and sets the `primary_script` field of the METADATA to the primary script of the font.


RosaWagner:
I have this error

```
(env3) MacBook-de-Rosalie:fonts rosalie$ gftools set-primary-script ofl/ibmplexsansdevanagari
  File "/Users/rosalie/Google/env3/bin/gftools-set-primary-script.py", line 55, in <module>
    main(args.directory)
  File "/Users/rosalie/Google/env3/bin/gftools-set-primary-script.py", line 37, in main
    script = primary_script(TTFont(first_font), ignore_latin=False)
  File "/Users/rosalie/Google/env3/lib/python3.9/site-packages/gftools/utils.py", line 530, in primary_script
    g = classifyGlyphs(lambda uv:list(ftunicodedata.script_extension(chr(uv))), ttFont.getBestCmap(), gsub=ttFont.get("GSUB"))
  File "/Users/rosalie/Google/env3/lib/python3.9/site-packages/ufo2ft/util.py", line 284, in classifyGlyphs
    glyphSets.setdefault(key, set()).add(glyphName)
TypeError: unhashable type: 'list'
```

simoncozens:
You have an old ufo2ft.

RosaWagner:
@simoncozens doesn't seem to work on [Tiro Indic fonts](https://fonts.google.com/?query=Tiro) (except Tiro Telugu) + Some IBM Plex (hebrew, thai and thai looped)

simoncozens:
This is hard to know how to detect. For example, Tiro Tamil contains 168 Tamil glyphs and 723 Latin glyphs. It's definitely a Latin font with a subsidiary Tamil. :-)

RosaWagner:
I guess if the font name contains the script name, we can assume the primary script?

RosaWagner:
The script is setting "Hani" for all CJK families, but it doesn't seem to work in sandbox. I gonna test "Jpan" for Japanese instead

RosaWagner:
As a result of my test,

Japanese fonts (full set) : "Jpan"
Hangul: "Kore"

For now we don't have traditional chinese families other than Noto, but I would expect "Hant". I tested "Hans" in simplified chinese fonts but dev-sandbox is not updating, so I don't know.

"Hina" (for SC and TC), "Hira" (for Hiragana) and "Kana" (Katakana only fonts) don't work.

simoncozens:
@m4rc1e I'd still like this to be merged, especially since primary_script is now becoming more of a thing.

yanone:
+1 for merging this. I'd like to implement a fontbakery check for primary_script and wouldn't want to duplicate the code, rather use gftools directly to generate primary_script and see whether it's missing from METADATA.pb

Lib/gftools/builder/autohint.py
```diff
@@ -65,24 +64,11 @@
...
-    ):
-        return
-
-    most_common = script_count.most_common(1)
-    if most_common:
-        script = most_common[0][0].lower()
-        if script in AUTOHINT_SCRIPTS:
-            return script
+def autohint_script_tag(ttFont, discount_latin=False):
+    script = primary_script(ttFont, discount_latin=discount_latin)
```
yanone, line 68: The parameter is called `ignore_latin` in the method definition

yanone:
I tested the bare `primary_script()` method on a number of Arabic and South Asian fonts and found it to work reasonably well. It only failed to produce a primary script for [Alkatra](https://github.com/yanone/Alkatra/blob/master/fonts/variable/Alkatra%5Bwght%5D.ttf) because it contains a total of four scripts incl. Latin, which I guess is expected after studying the code.

yanone:
I have another wish:

Can a third result type be added to `primary_script()` where it determines that no script comes out as a clear winner, but the overwhelming majority of glyphs are non-Latin?

I want to use this method directly in the upcoming related fontbakery check to avoid code duplication, and I wouldn't want it to fail silently (`primary_script()` returning `None`) when it’s clearly a non-Latin font, like Alkatra (linked above), which supports three South Asian scripts in addition to Latin.

yanone:
Update: I just realized that `gftools` is not a dependence of `fontbakery`, and I’m guessing you want it to stay like that, so I would have to duplicate the code into `fontbakery` anyway. So never mind, I can probably add my desired feature myself then.

m4rc1e:
I don't mind getting this across the line today.

I'm wondering, do we expect every new family to use the primary_script field? If so, I think we need to add it `gftools add-font` as well.

RosaWagner:
Only the new families that are non-LCG

m4rc1e:
Awesome thanks. I've just hooked this up to `gftools add-font`. If the family's metadata doesn't have a primary_script field, it'll add one if the primary isn't Latin, Greek or Cyrillic.

PR #629, opened 2022-11-17
Sync fonts_public.proto with internal

simoncozens:
This gives us the minisite_url field, making sure we don't overwrite it when updating things with the packager.

### DIFF
--- a/Lib/gftools/fonts_public.proto
+++ b/Lib/gftools/fonts_public.proto
@@ -24,13 +24,16 @@ message FamilyProto {
   optional bool is_noto = 13;
   repeated string languages = 14;
   repeated FamilyFallbackProto fallbacks = 15;
-  repeated GlyphGroupProto sample_glyphs = 16;
+  map<string, string> sample_glyphs = 16;
   optional SampleTextProto sample_text = 17;
   optional string display_name = 18;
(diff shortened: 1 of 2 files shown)


Lib/gftools/fonts_public.proto
```diff
@@ -81,55 +84,6 @@ message FamilyFallbackProto {
   // Next = 6
 }
 
-//
```
m4rc1e, line 84: I wonder if it's safe to strip this out? don't some Noto projects depend on these fields? e,g https://github.com/google/fonts/blob/b17d6bf90cadd249bb7a5135f3d76c23a2c89663/ofl/notoserifethiopic/METADATA.pb#L34

m4rc1e: I'd be more comfortable if we just added the `minisite_url` field only.

m4rc1e: Argh, it didn't show the block I highlighted. My concern is you're also deleting all the lang metadata fields when I suspect they may be used for Noto.

simoncozens: No, what you're looking at is field 14 (`repeated string languages`).

I feel like it's better to minimise the changes between our copy and theirs, and theirs doesn't have the lang metadata (it belongs in the lang repo).

PR #635, opened 2022-11-28
Various builder enhancements

### LINKED ISSUE #293: [builder] default config filename
Per "convention over configuration" I propose builder should work without any args if a config.y[a]ml exists in the pwd. So these would be equivalent

    gftools builder config.yml 
    gftools builder config.yaml 
    gftools builder

### LINKED ISSUE #344: Builder: Add docs
gftools builder should have docs which are hosted in our [docs/](https://github.com/googlefonts/gftools/tree/main/docs) dir. Most designers don't open source code. Simplest solution is to just use the builder's doc string.

### LINKED ISSUE #350: Builder - woff2 variable font request
It would be useful to have an option of outputting variable fonts in woff2

### LINKED ISSUE #409: builder: Set SOURCE_DATE_EPOCH
See https://github.com/googlefonts/Unified-Font-Repository/issues/34

### LINKED ISSUE #472: [Builder] can't build non-interpolable statics fonts while fontmake does
I have this error when I want to build a font (has 3 masters, but is not variable) with gftools builder, but I tried to export it with only Fontmake and then it worked.

    raise FontmakeError(fontmake.errors.FontmakeError: In 'CrimsonText.glyphs' -> 'master_ufo/CrimsonText.designspace': Interpolating instance 'SemiBold' failed.: Failed to generate instance of glyph 'brevecomb_hookabovecomb.case': list index out of range. (Note: the most common cause for an error here is that the glyph outlines are not point-for-point compatible or have the same starting point or are in the same order in all masters.)

Source file + config.yaml:
[CrimsonText.zip](https://github.com/googlefonts/gftools/files/7959082/CrimsonText.zip)

### LINKED ISSUE #624: [builder] stops printing INFO messages after updating fontmake to 3.5.0
## Observed behavior

When using the `gftools builder config.yml` command, loglevel INFO messages are printed informing about process status. 
E.g. `INFO:GFBuilder:Building variable completed. All done!`

However, as soon as fontmake is updated to `fontmake==3.5.0`, builder (using same above command) stops printing the INFO messages.

## Expected behavior

To continue receiving the INFO messages for process status when using Fontmake 3.5

## Resources and exact process needed to replicate

- Create a new venv with the latest tools versions (fontmake-3.5.0 fonttools-4.38.0 glyphsLib-6.1.0 ufo2ft-2.29.0) and run the `gftools builder config.yml` command.
- Alternatively, just update fontmake in a previous venv and run the same builder command

### LINKED ISSUE #634: Builder assumes buildVariable is true instead of being smart
Continuation of https://github.com/googlefonts/gftools/issues/293

simoncozens:
- Try to determine config file location, fixes #293
- Configure basic logging, fixes #624
- Automatically determine whether to build variable or static, fixes #634
- Only interpolate statics if we have a variable font, since they're likely to be interpolatable. Fixes #472
- Set head.modified timestamp to that of source file, fixes #409
- Allow builder to change version and add gftools version
- Support WOFF2 variable fonts, fixes #350
- Add docs and script to make them, fixes #344


simoncozens:
This will now need much rebasing onto main...

m4rc1e:
I'll make a coffee and rebase this.

PR #637, opened 2022-11-28
Skip directories from the push lists if their parent is also present

### LINKED ISSUE #589: gen-push-list: avoid duplicated path
Pushing Noto and knowledge will generate this kind of list;

ofl/notosanstamil
ofl/notosanstamil/article

cc-by-sa/knowledge/glossary/terms/grade_axis
cc-by-sa/knowledge/glossary/terms/grade_axis/images

I guess we don't need the path with article or images, since the one just before covers that.

simoncozens:
- Add path transformations which work for "scripts" as well as "languages"
- Skip subdirectories when parent is already in push list. Fixes #589


simoncozens:
Note this contains a more general fix to path transformations than the recently merged 8b25af9.

m4rc1e:
Nice and thank you very much!

PR #633, opened 2022-11-27
Remove list-variable-source

simoncozens:
This script only really worked on Marc's computer, and these days people doing multiple-master stuff in Glyphs expect to go to variable fonts, so it's not really a thing any more. Fixes #9.

Issue #560, opened 2022-05-16
gftools builder: [WinError 193] error


dy:
Trying to follow [the docs](https://rosaliewagner.notion.site/Build-the-fonts-1e6bd5b5df10418d8b7e447cf50ec72e), installed latest python and pip, installed gftools. 

On attempt to run `gftools builder --help` getting:
  
```
  File "C:\Users\dfcre\AppData\Local\Programs\Python\Python310-32\Scripts\gftools", line 99, in <module>
    p = subprocess.Popen(args, stdout=sys.stdout,
  File "C:\Users\dfcre\AppData\Local\Programs\Python\Python310-32\lib\subprocess.py", line 966, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Users\dfcre\AppData\Local\Programs\Python\Python310-32\lib\subprocess.py", line 1435, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
OSError: [WinError 193] %1 is not a valid Win32 application
```

Please help.

simoncozens:
Fixed by #638.

Issue #531, opened 2022-03-30
Inconsistent use of absl flags vs argparse


rsheeter:
We apparently use argparse and absl flags, try `grep argparse bin/*.py`, `grep absl.*flags bin/*.py`. Pick one. In nanoemoji and other recent tools we use absl, but either is fine.

simoncozens:
Fixed in #638 except for a few scripts which need deprecating anyway.

Issue #390, opened 2021-07-08
[wish] Refactor, move more advanced logic from `bin` scripts to Lib


twardoch:
@m4rc1e @felipesanches 

GFTools is a collection of somewhat random scripts. Many of them are useful, but above all, many of them contain code that could be useful for other purposes. As an example, [gftools-udhr.py](https://github.com/googlefonts/gftools/blob/main/bin/gftools-udhr.py) contains a useful parser for UDHR XMLs, and it’d be good to move the classes to a separate module in the lib, and make the CLI tool very shallow. 

For example, the classes could be in `https://github.com/googlefonts/gftools/tree/main/Lib/gftools/udhr.py` and the script could just import that (for backwards compatibility, it could do `from gftools.udhr import *`) and then do very minimal logic. Well, in fact that particular script does not really have any CLI stuff at all :) 

What do you think?

felipesanches:
sounds good

m4rc1e:
Agreed.

The fix scripts already have their own module, https://github.com/googlefonts/gftools/blob/main/Lib/gftools/fix.py

simoncozens:
Fixed by #638.

PR #638, opened 2022-11-29
Refactor binaries into packages

simoncozens:
This is a big PR. It:

* Moves the build system to setuptools+pyproject.toml, fixing #559.
* Moves all the bin/gftools-* scripts into a modules under a `gftools.scripts` package, and setting them as endpoints installed by setuptools, fixing #390 and probably also fixing #560.
* Refactors *most* of the absl-based scripts to use argparse, fixing #531.
* Deprecates a few scripts (font-diff and space-check) and removes a few obviously broken ones (gftools-cldr, gftools-list-variable-source, gftools-udhr), fixing #9 and #14 and closing #633.
* Moves the tests from Lib/gftools/tests to tests/, since they don't need to be installed on the users' system.
* Updates tests so that they *actually* begin to test some of the scripts' functionality.
* Fixes an issue where the builder tests had to be run separately.

### DIFF
--- a/Lib/gftools/html.py
+++ b/Lib/gftools/html.py
@@ -13,6 +13,7 @@ import time
 from copy import copy
 import shutil
 from collections import namedtuple
+import ntpath
 from gftools.utils import (
     font_sample_text,
     download_file,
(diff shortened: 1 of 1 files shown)


m4rc1e:
Overall looks good but I have a lot of beef with poetry. What does it bring to the table that pip doesn't?

Installing on my system isn't working due kurbopy which needs cargo. Which package uses kurbopy?

simoncozens:
There's no poetry here.

simoncozens:
kurbopy is now used by collidoscope which is used by fontbakery. And I think there's another problem with collidoscope which now prevents it being used in fontbakery, ugh.

m4rc1e:
Thanks.

Sorry, I thought pyproject.toml was a poetry thing, didn't realise it was established in pep 518.

m4rc1e (approved):
Thanks Simon. this is great!

PR #641, opened 2022-12-01
add-axis: do not include fallbacks

m4rc1e:
We've decided that we won't be adding fallbacks to custom axes anymore. We only need to declare a single default axis value going forward.


vv-monsalve:
This is the rationale per the Axis Registration Protocol:

> A fallback at the axis default value is required. For server implementation reasons, new custom axis registries require to include one single fallback. It must be called “Default” (reserving the use of  “Regular” for `wght` and “Normal” for `wdth` axes) and the value should match the axis “default_value”

PR #648, opened 2023-01-11
Stop removing the MVAR table

### LINKED ISSUE #580: Don't kill MVAR table in fix-font
We've re-thought the fontbakery check that complained about the MVAR table (https://github.com/googlefonts/fontbakery/issues/3400) but we still drop the MVAR table in fix-font. The problem with this is that fonts instantiated from a VF with no MVAR table have the wrong vertical metrics.

madig:
Closes #580.

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -84,7 +84,6 @@ UNWANTED_TABLES = frozenset(
         "TSI3",
         "TSI5",
         "prop",
-        "MVAR",
         "Debg",
     ]
 )
@@ -613,7 +612,6 @@ def fix_font(font, include_source_fixes=False, new_family_name=None, fvar_instan
         fix_unhinted_font(font)
 
     if "fvar" in font:
-        remove_tables(font, ["MVAR"])
         name_table = font["name"]
         variation_ps_name = name_table.getName(25, 3, 1, 0x409)
         if not variation_ps_name:


Lib/gftools/fix.py
```diff
@@ -613,7 +612,6 @@ def fix_font(font, include_source_fixes=False, new_family_name=None, fvar_instan
         fix_unhinted_font(font)
 
     if "fvar" in font:
-        remove_tables(font, ["MVAR"])
```
m4rc1e, line 616: By dropping this, we won't remove the other redundant tables. Would prefer

```PYTHON
remove_tables(font)
```

m4rc1e: Sorry, you are correct. This func already gets called earlier.

m4rc1e (approved):

m4rc1e:
I'll chat to Chris before I merge this. I may need to update fontbakery as well.

chrissimpkins:
Marc and I discussed this today.  OK to merge.  Note that we will begin to add MVAR tables to new VF families and existing VF families that update their gftools dependency version as of the release that includes these changes.

fontbakery was changed to stop failing VF files that contain MVAR in https://github.com/googlefonts/fontbakery/commit/115b28575f9bf48f45a089efebdd018da5be3123.  As I understand it, there is no longer a check for MVAR at all in fontbakery.  There will be no WARN or FAIL for VFs with MVAR.

PR #657, opened 2023-02-21
Add option to interpolate between masters for static fonts.

bramstein:
We recently moved to gftools v0.9.23 and noticed some of our Glyphs-based families stopped building correctly. Instead generating static fonts for each export in Glyphs, only the masters were built.

I believe the change in https://github.com/googlefonts/gftools/commit/bfb90fadbfde4431828f80b531f202da4a64eb0d is the cause. We have quite a few families that are interpolatable but not variable. That means that glyphs can be interpolated between some masters, but not all masters, so enabling the creation of variable fonts (through `buildVariable`) wouldn't help. I have attached an example Glyphs file (excuse my terrible design skills) of a family with some compatible masters using interpolation.

[Example.zip](https://github.com/googlefonts/gftools/files/10793928/Example.zip)

We would like to add a new `gftools builder` option that allows us to explicitly opt into interpolation in these cases.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -107,6 +107,7 @@ required, all others have sensible defaults:
 * ``expandFeaturesToInstances``: Resolve all includes in the sources' features, so that generated instances can be compiled without errors. Defaults to ``true``.
 * ``reverseOutlineDirection``: Reverse the outline direction when compiling TTFs (no effect for OTFs). Defaults to fontmake's default.
 * ``removeOutlineOverlaps``: Remove overlaps when compiling fonts. Defaults to fontmake's default.
+* ``interpolateMasters``: Interpolate between masters when generating static fonts. Defaults to ``false``.
 """
 
 from fontmake.font_project import FontProject
@@ -310,6 +311,8 @@ class GFBuilder:
             self.config["addGftoolsVersion"] = True
         if "decomposeTransformedComponents" not in self.config:
             self.config["decomposeTransformedComponents"] = True
+        if "interpolateMasters" not in self.config:
+            self.config["interpolateMasters"] = False
 
     def build_variable(self):
         self.mkdir(self.config["vfDir"], clean=True)
@@ -498,7 +501,7 @@ class GFBuilder:
                 "output_dir": directory,
                 "optimize_cff": CFFOptimization.SUBROUTINIZE,
             }
-            if self.config["buildVariable"]:
+            if self.config["buildVariable"] or self.config["interpolateMasters"]:
                 args["interpolate"] = True
             self.logger.info("Creating static fonts from %s" % source)
             for fontfile in self.run_fontmake(source, args):
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -501,7 +501,7 @@ class GFBuilder:
                 "output_dir": directory,
                 "optimize_cff": CFFOptimization.SUBROUTINIZE,
             }
-            if self.config["buildVariable"] or self.config["interpolateMasters"]:
+            if self.config["buildVariable"] or self.config["interpolate"]:
                 args["interpolate"] = True
             self.logger.info("Creating static fonts from %s" % source)
             for fontfile in self.run_fontmake(source, args):
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -107,7 +107,7 @@ required, all others have sensible defaults:
 * ``expandFeaturesToInstances``: Resolve all includes in the sources' features, so that generated instances can be compiled without errors. Defaults to ``true``.
 * ``reverseOutlineDirection``: Reverse the outline direction when compiling TTFs (no effect for OTFs). Defaults to fontmake's default.
 * ``removeOutlineOverlaps``: Remove overlaps when compiling fonts. Defaults to fontmake's default.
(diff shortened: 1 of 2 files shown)


IvanUkhov:
It might have been discussed before, but would it not make it more scalable to introduce a configuration option for propagating arbitrary arguments to `fontmake`? There could be many different custom use cases, past and future, and all of them could be tackled with such a change once and for all.

```python
    def build_a_static_format(self, format, directory, postprocessor):
        self.mkdir(directory, clean=True)
            args = {
                "output": [format],
                "output_dir": directory,
                "optimize_cff": CFFOptimization.SUBROUTINIZE,
            }
            if self.config["buildVariable"]:
                args["interpolate"] = True
            args = {**args, **self.config["fontmake_args"]} # Override and enrich
            self.logger.info("Creating static fonts from %s" % source)
            for fontfile in self.run_fontmake(source, args):
                self.logger.info("Created static font %s" % fontfile)
                postprocessor(fontfile)
                self.outputs.add(fontfile)
```

anthrotype:
> interpolatable but not variable

at first this didn't make sense to me, because these two adjective should be synonyms.. but then I remembered that fontmake has this -i/--interpolate option which controls whether to build instances, and if not set it only builds the masters. Instances usually are interpolated between masters, but when they sit exactly on the same location as one of the masters they are not actually "interpolated", but simplied copied from the respective master plus they get all the custom parameters that are associated with that instance. If all the instances are equal to one of the masters, then no interpolation occurs and the masters may well have different incompatible designs; 'instances' in this case are just containers of custom parameters.
Arguably that option should be true by default, especially since from Glyphs.app one can only export instances, never masters as such; the current default is probably a legacy of some old build pipeline (probably some Noto fonts expected that option to not be set).

bramstein:
@IvanUkhov: That's a good idea :+1: 

@anthrotype: Thanks for taking a look! I agree the option should be true by default. I didn't make it the default because the change in https://github.com/googlefonts/gftools/commit/bfb90fadbfde4431828f80b531f202da4a64eb0d looked intentional.

m4rc1e:
Interestingly, I can generate them individually by doing:

```
fontmake -g Example.glyphs -i "Light" -o ttf
fontmake -g Example.glyphs -i "SemiBold" -o ttf
fontmake -g Example.glyphs -i "Regular" -o ttf
```

m4rc1e:
Ok, I think I understand the issue now.

Emma, Simon and Rosalie encountered sources which contain several masters but the masters in these sources are **not** compatible. They also have an instance/export for every master.

Bram has sources where some of the masters are compatible and other aren't. There are more instances/exports than masters.

Perhaps the cleanest solution is to allow everyone to just set the "interpolate" argument directly?

Lib/gftools/builder/schema.py
```diff
@@ -82,6 +82,7 @@
         Optional("googleFonts"): Bool(),
         Optional("category"): UniqueSeq(Enum(CATEGORIES)),
         Optional("reverseOutlineDirection"): Bool(),
+        Optional("interpolateMasters"): Bool(),
```
m4rc1e, line 85:
```suggestion
        Optional("interpolate"): Bool(),
```

Lib/gftools/builder/__init__.py
```diff
@@ -498,7 +501,7 @@ def build_a_static_format(self, format, directory, postprocessor):
                 "output_dir": directory,
                 "optimize_cff": CFFOptimization.SUBROUTINIZE,
             }
-            if self.config["buildVariable"]:
+            if self.config["buildVariable"] or self.config["interpolateMasters"]:
```
m4rc1e, line 504:
```suggestion
            if self.config["buildVariable"] or self.config["interpolate"]:
```

Lib/gftools/builder/__init__.py
```diff
@@ -310,6 +311,8 @@ def fill_config_defaults(self):
             self.config["addGftoolsVersion"] = True
         if "decomposeTransformedComponents" not in self.config:
             self.config["decomposeTransformedComponents"] = True
+        if "interpolateMasters" not in self.config:
+            self.config["interpolateMasters"] = False
```
m4rc1e, line 315:
```suggestion
        if "interpolate" not in self.config:
            self.config["interpolate"] = False
```

Lib/gftools/builder/__init__.py
```diff
@@ -107,6 +107,7 @@
 * ``expandFeaturesToInstances``: Resolve all includes in the sources' features, so that generated instances can be compiled without errors. Defaults to ``true``.
 * ``reverseOutlineDirection``: Reverse the outline direction when compiling TTFs (no effect for OTFs). Defaults to fontmake's default.
 * ``removeOutlineOverlaps``: Remove overlaps when compiling fonts. Defaults to fontmake's default.
+* ``interpolateMasters``: Interpolate between masters when generating static fonts. Defaults to ``false``.
```
m4rc1e, line 110:
```suggestion
* ``interpolate``: Enable fontmake --interpolate flag. Defaults to ``false``.
```

m4rc1e:
@bramstein I've made some suggestions so that your code directly sets the fontmake interpolate argument. I hope this will solve your issue.

bramstein:
@m4rc1e Sounds good to me, it will definitely solve my issue. Thanks for the suggestions!

m4rc1e:
@bramstein https://github.com/googlefonts/gftools/releases/tag/v0.9.24 includes your fix. Thanks!

PR #659, opened 2023-02-23
fix: return a new TTFont instance when running fix_font and fix_family

m4rc1e:
Atm, if you use `fix_font` or `fix_family`, fixes are applied in place. Unfortunately, the COLR fixes won't be applied since `fix_colr_v1_add_svg` needs to return a new TTFont instance since we're calling an external tool using subprocess. This PR changes `fix_family` and `fix_font` so they don't fix inplace but return TTFont instances instead.

cc @yanone


yanone:
This works

simoncozens (approved):
LGTM!

PR #662, opened 2023-03-03
builder: add arg to enable or disable fontmake's check-compatibility

### LINKED ISSUE #660: [builder] Making more (or all) fontmake arguments available
After adding alternate layers for swapping cursive shapes, the build process for Geologica fails with the error "Compatibility check failed". Fontmake has the `--no-check-compatibility` argument that works to build the font ok, but it's not yet available.

Only some [Fontmake arguments are exposed](https://github.com/googlefonts/gftools/pull/617) to be used by Builder. But it would be of great help to have the option to make all the already available/in-use fontmake arguments to be able to build from Builder and keep using its facilities.

cc @m4rc1e

m4rc1e:
Partly fixes #660 by allowing Viv and myself to disable Fontmake's compatibility checker.

Allowing the builder to take arbitrary args is tricky since we currently check if a user's config file has the correct args and they're also the correct type.


simoncozens:
This looks good to me. I'm starting to think that at some point we should refactor fill_config_defaults into a dictionary merge, at least for the "static" keys: `self.config = {**CONFIG_DEFAULT, **self.config}`

simoncozens (approved):

vv-monsalve:
Brilliant! Thank you, Marc

PR #664, opened 2023-03-05
Add option to interpolate with MutatorMath (for extrapolation, anisotropic locations)

moyogo:
This add the useMutatorMath paramater to call fontmake with `--use-mutatormath` for extrapolation or anisotropic locations when generating instances.


m4rc1e:
Great! thanks for ensuring this only runs for statics as well!

PR #670, opened 2023-03-23
add builder checks for check_compatibility

m4rc1e:
@bramstein Apologies for breaking the builder. Since you folks are now relying on it as well, I'm going to write some unit tests so we don't break anything going forward. The test in this PR ensures that your PR works. I plan to add more going forward.


bramstein:
@m4rc1e That's great, thanks! We have very similar tests internally that run through CI whenever we update our foundry template repository. That's how we caught this quite early :smile:

PR #689, opened 2023-06-29
add greendot algo

m4rc1e:
Still WIP.

@RosaWagner @emmamarichal to run this tool use the following command `python -m gftools.greendot min_opsz,max_opsz` e.g for Imbue `python -m gftools.greendot 10,100`.

So far my code passes all the examples listed in the greendot doc. However, I still need to clean up and check more. I think I'm missing some edgecases.

### DIFF
--- a/Lib/gftools/greendot.py
+++ b/Lib/gftools/greendot.py
@@ -1,0 +1,80 @@
+"""
+David Berlow's Green Dot algorithm:
+https://docs.google.com/document/d/15652Yabs0prnocpjG1TxG6zrfFSIYqOwlSIMkdbgqfg/edit?resourcekey=0-DXNZQLV2TbSqyn9HCLfhFA
+"""
+from collections import OrderedDict
+import sys
+
+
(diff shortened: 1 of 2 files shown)


m4rc1e:
It probably makes more sense to include this in the axisregistry.

emmamarichal:
Just tested it on Bricolage project, it's super useful, thanks a lot!!

graphicore:
I tried to read the referenced document: https://docs.google.com/document/d/15652Yabs0prnocpjG1TxG6zrfFSIYqOwlSIMkdbgqfg/edit?resourcekey=0-DXNZQLV2TbSqyn9HCLfhFA but it is not public. It should either be accessible or the inline documentation should be fleshed out.

vv-monsalve:
Also tested for Illusio. It has an 8-144 range and the tool throw: `[14, 28, 36, 72, 120]`

m4rc1e:
Closing since I've moved this to the axisregistry.

PR #696, opened 2023-07-23
Add support for specifying custom GlyphData XML files

khaledhosny:
To be used with Glyphs sources.

(this is #694 but for `noto` branch as I need this in notobuilder).


m4rc1e:
@khaledhosny looks like a dup of #694?

khaledhosny:
> @khaledhosny looks like a dup of #694?

This is against the `noto` branch not `main`. I need to use in [notobuilder](https://github.com/notofonts/notobuilder) (which [uses `noto` branch](https://github.com/notofonts/notobuilder/blob/5cdd86efd2d04368a3033dbdc80947100589f6cf/setup.py#L17) instrad of released version of gftools).

cc @simoncozens

Issue #693, opened 2023-07-20
[packager] stop removing custom sample text


RosaWagner:
such as in https://github.com/google/fonts/pull/5281#discussion_r1269344202

PR #699, opened 2023-07-28
stop packager from deleting minisite_url from metadata.pb

RosaWagner:
Saw same issue as in #693 so tried same technique


m4rc1e (approved):
Thanks for opening this!

PR #705, opened 2023-08-07
Provide UFO structure

simoncozens:
Fonttools v3.7.0 changed the API (thanks!) requiring an extra keyword argument to `run_from_glyphs`. See https://github.com/googlefonts/googlefonts-project-template/issues/109

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -389,6 +389,9 @@ class GFBuilder:
             args["check_compatibility"] = self.config["checkCompatibility"]
             if "glyphData" in self.config:
                 args["glyph_data"] = self.config["glyphData"]
+            # The following line should be removed once we require
+            # fontmake >= 3.7.1
+            args["ufo_structure"] = "package"
             FontProject().run_from_glyphs(source, **args)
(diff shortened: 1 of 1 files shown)


Lib/gftools/builder/__init__.py
```diff
@@ -389,6 +389,9 @@ def run_fontmake(self, source, args):
             args["check_compatibility"] = self.config["checkCompatibility"]
             if "glyphData" in self.config:
                 args["glyph_data"] = self.config["glyphData"]
+            # The following line should be removed once we require
+            # fontmake >= 3.7.1
+            args["ufo_structure"] = "package"
```
anthrotype, line 394: `ufo_structure` parameter didn't exist before 3.7.0 so in practice passing this parameter is equivalent to requiring fontmake >= 3.7.0 (which you did not do in pyproject.toml's dependencies as far as I can tell).
It's probably better to do nothing and rely on the latest 3.7.1 to just fix the issue.
I can also "yank" 3.7.0 if you want (make it unavailable from pip install unless one explicitly requests fontmake==3.7.0), but maybe it's not worth it, it was online only a couple days

PR #709, opened 2023-09-01
Make the qa tool more robust

simoncozens:
This improves the robustness of ghtools.qa in a couple of ways:

- Ensures that the fontbakery report is posted as a GH comment just after fontbakery runs, so it is encapsulated from other methods.
- Wraps each function in a decorator which catches exceptions so that an exception in one test does not crash the whole tool.
- Posts a notification of any exceptions/failures to GH.

This is split into fairly small self-contained commits, so may be easiest to review one commit at a time.

### DIFF
--- a/Lib/gftools/qa.py
+++ b/Lib/gftools/qa.py
@@ -40,7 +40,7 @@ class FontQA:
             diffenator=True,
             diffbrowsers=False,
         )
-    
+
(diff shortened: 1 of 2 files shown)


felipesanches (approved):
ok. That's all looking great!

simoncozens:
Cool, thank you!

PR #713, opened 2023-09-01
Stop duplicating empty sample text

### LINKED ISSUE #707: [Packager] should not add empty fields in metadata.pb
In [the PR for Ubuntu](https://github.com/google/fonts/pull/6609/files#diff-910688eb24ef99919a229972cd8fdfc7e0fbb70e1c31904b43afb2dc2c85de0c), a sample text entry was add but empty. I don't think this entry should be added automatically if there was not one already before

Also this PR from Rod to correct Wixmadefor Display: https://github.com/google/fonts/pull/6646

This issue seems to only affect Marc's PRs

simoncozens:
Fixes #707

Because `sample_text` is a Protobuf "message" (i.e. a nested dictionary), unlike a Python dictionary and unlike a standard scalar field, this has a boolean value of true even if it has no fields. We have to check if there's actually anything in it.

### DIFF
--- a/Lib/gftools/scripts/add_font.py
+++ b/Lib/gftools/scripts/add_font.py
@@ -150,7 +150,7 @@ def _MakeMetadata(args, is_new):
       metadata.display_name = old_metadata.display_name
     if old_metadata.primary_script:
       metadata.primary_script = old_metadata.primary_script
-    if old_metadata.sample_text:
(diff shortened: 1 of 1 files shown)


felipesanches:
I think you assigned this to me by mistake. You probably wanted to add me as a code-reviewer instead.

felipesanches (approved):
yes, that's good!

PR #718, opened 2023-09-05
Replace hyperglot

### LINKED ISSUE #498: hyperglot is a GPLv3 dependency
see: https://github.com/felipesanches/gflanguages/pull/4

moyogo:
- scripts/lang.py is not needed, or belongs in https://github.com/googlefonts/lang.
- Replacing hyperglot.parse and removing the dependency, fixes #498.
  The logic should probably be in a function in https://github.com/googlefonts/lang itself, or shaperglot to should be used instead.


simoncozens:
Thanks for this. I think I would prefer to first move that logic to parse bases into gflanguages; gflanguages should be responsible for turning its own protobuf data into something more Python-friendly. Then the new method can be used here and in shaperglot.

Issue #719, opened 2023-09-08
Ninja Builder / deleted files


emmamarichal:
Hi!

With the last GFtools update (3.4), some files are not deleted after exporting fonts:
- build.ninja
- .ninja_log (hidden)
- instance_ufo
- master_ufo
Same when I run gftools QA.

If I don't delete them manually and if I want to re-run a build and QA, it causes this issue:
<img width="842" alt="Capture d’écran 2023-09-08 à 10 24 54" src="https://github.com/googlefonts/gftools/assets/64773544/53c0f363-5f0a-408f-bcf4-92feded73c09">

Is it possible to add an automatic deletion as before?
Thanks! :)

eliheuer:
Thanks, I made a PR here fixing this: https://github.com/googlefonts/gftools/pull/721

It might not get merged right away, but it's a pretty simple fix and if you install gftools in editable mode `cd gftools && pip install -e .` you just need to edit `gftools/Lib/gftools/builder/_ninja.py` with the [changes I made](https://github.com/googlefonts/gftools/pull/721/files).

eliheuer:
@emmamarichal PR was merged, just install gftools in editable mode or wait for the next release.

emmamarichal:
great, thank you @eliheuer!

emmamarichal:
@eliheuer oh, is it possible to do the same for gftools qa? Or should I open an issue somewhere else (or re-open this one)? Don't know if it's related to diffenator or gftools qa

eliheuer:
@emmamarichal a new issue would be helpful, I'm looking at this now along with https://github.com/googlefonts/gftools/issues/728 but more context might help.

emmamarichal:
@eliheuer 
When I run `gftools qa -f *.ttf -a -gfb` I get these files: 
- folder "out", with fontbakery, diffenator, diffbrowers, fonts, fonts before
- build.ninja
- .ninja.log (hidden)

All of these were usually deleted when I build the fonts again. 
But actually, I noticed that build.ninja and .ninja.log are deleted, so maybe it's intentional to keep only the previous "out" file even if we build the fonts again.

So maybe you can ignore this issue eventually

PR #721, opened 2023-09-10
Clean up build files after Ninja Builder is done

eliheuer:
This fixes issue https://github.com/googlefonts/gftools/issues/719

### DIFF
--- a/Lib/gftools/builder/_ninja.py
+++ b/Lib/gftools/builder/_ninja.py
@@ -7,6 +7,7 @@ import sys
 import ufoLib2
 import yaml
 import os
+import shutil
 from gftools.builder import GFBuilder
 from fontTools.designspaceLib import (
     DesignSpaceDocument,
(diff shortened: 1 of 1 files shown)


Lib/gftools/builder/_ninja.py
```diff
@@ -61,6 +62,23 @@ def build(self):
...
+        # Clean up temp build files
+        search_directory = os.getcwd()
+        target_names = ["build.ninja", ".ninja_log", "instance_ufo", "master_ufo"]
+        for root, dirs, files in os.walk(search_directory, topdown=False):
+            for file in files:
+                if file in target_names:
+                    file_path = os.path.join(root, file)
+                    os.remove(file_path)
+                    print(f"Removed file: {file_path}")
+        for dir_name in dirs:
```
simoncozens, line 74: Shouldn't this be within the "os.walk" loop?

simoncozens (changes requested):

Lib/gftools/builder/_ninja.py
eliheuer: Sure, like this? I tested this with a few different font repos and it seems to work, and pushed it as an update to this PR (e558616).
```python
        # Clean up temp build files
        search_directory = os.getcwd()
        target_names = ["build.ninja", ".ninja_log", "instance_ufo", "master_ufo"]
        for root, dirs, files in os.walk(search_directory, topdown=False):
            for file in files:
                if file in target_names:
                    file_path = os.path.join(root, file)
                    os.remove(file_path)
                    print(f"Removed file: {file_path}")
            for dir_name in dirs[:]:
                if dir_name in target_names:
                    dir_path = os.path.join(root, dir_name)
                    shutil.rmtree(dir_path)
                    print(f"Removed directory: {dir_path}")
```

eliheuer: The shallow copy of dirs `dirs[:]` was ChatGPTs idea, but I looked into it and it made sense to me. ChatGPTs explanation:
>By creating a copy with dirs[:], you can safely remove elements from the original dirs list within the loop without affecting the ongoing iteration. This is particularly useful when you want to conditionally remove items from a list while iterating over it.

simoncozens: Nice, thanks.

PR #725, opened 2023-09-14
Update add_designer.py

emmamarichal:
@m4rc1e 
In the spreadsheet, the "Name" column became "Designer Name" instead of "Name", we need to update it in the script.
I hope I changed the right line :)


vv-monsalve:
Hi Emma, I changed the column name in the Spreadsheet back to "Name" for it to work okay with the add-designer tool, and it was working the last time I used it (last week). 
If you are using a local `.xlsx` file you can change the column name there and it would work ok. Is there any specific purpose for the column to be named "Designer Name" in the spreadsheet?

vv-monsalve:
Fair enough :)

m4rc1e (approved):

Issue #740, opened 2023-09-29
[builder] Calling UFOBuilder directly doesn't set minimal=True and thus prevents color layer processing


yanone:
Since the ninja builder came about, `gftools` relies on its own routine to generate designspace objects from Glyphs sources, rather than having `fontmake` do that.

`fontmake` however calls `UFOBuilder()` with `minimal=True`, [which the ninja builder doesn't](https://github.com/googlefonts/gftools/blob/main/Lib/gftools/builder/_ninja.py#L150), and the effect of that is that
1. color layers aren't processed, which rely on `minimal=True`
2. possibly the UFO builder is overall slower than it should for one-way font generation

So `minimal=True` needs to be set explicitly (which is set to `False` by default) to ensure correct color layer handling in `glyphsLib`.

I'll attach a PR in a second.

anthrotype:
actually, the gftools ninja builder is still calling `fontmake -o ufo -g input.glyphs` command to convert from .glyphs to DS+ufo (see the `glyphs2ufo` ninja build rule). It seems to be using the UFOBuilder just to get a minimal designspace file in order to obtain the paths of the master UFOs that _will_ be generated when the `glyphs2ufo` build rule is actually run..

@simoncozens can you change glyphsLib to add APIs to obtain these pieces of info without you having to hack into the bowels of UFOBuilder (e.g. like overriding methods with `lambda: True` to make them noop, etc.)?

simoncozens:
> can you change glyphsLib to add APIs to obtain these pieces of info without you having to hack into the bowels of UFOBuild

No, I don't think I can. For example, if we skip the line

```
            list(self.masters)  # Make sure that the UFOs are built
```

then `to_designspace_sources` fails because `self._sources` is not filled in; that happens in `to_ufo_font_attributes`. glyphsLib.builders really isn't designed to work in a compartmentalised way, and I don't really think it's worth the effort of completely rearchitecturing it right now. I think under the circumstances hacking into the bowels of overriding methods is the least bad approach.

simoncozens:
> color layers aren't processed, which rely on minimal=True

That's kind of weird. `minimal=True` should do *less* work than `minimal=False`.

khaledhosny:
> > color layers aren't processed, which rely on minimal=True
> 
> That's kind of weird. `minimal=True` should do _less_ work than `minimal=False`.

[...]

PR #741, opened 2023-09-29
minimal=True explicitly

yanone:
Fixes https://github.com/googlefonts/gftools/issues/740


simoncozens:
This is now redundant with builder2.

PR #743, opened 2023-10-02
[builder2] Pass --glyph-data to fontmake when converting glyphs to ufo

khaledhosny:
This seems to have been dropped during the refactoring.


khaledhosny:
cc @simoncozens, without this my noto arabic builds are seriously broken.

Issue #750, opened 2023-10-05
building from quadratic with tt-hinting


RosaWagner:
Posted an issue in fontmake https://github.com/googlefonts/fontmake/issues/1045 but it may be an issue with the builder.

Cosimo says that to build a font from quadratic curves + hinting we should do this:

> for a font with quadratic sources and manual hinting, in addition to --keep-direction, you could also pass --ttf-curves=keep-quad so cu2qu is skipped as well (it should not modify the quads but it's quicker if you know in advance there are no cubic to convert). Also --keep-overlaps if you're building a static font, so the original quadratic outlines are not messed with.

simoncozens:
You can set:

```
reverseOutlineDirection: false
removeOutlineOverlaps: false
```

The "keep-quad" is just a speed-up. These two argument don't work in the ninja builder, which I'll fix in a moment, but in your case you are not using that because the ninja builder doesn't support vttSources so it falls back to the standard builder.

PR #756, opened 2023-10-12
Support a couple more fontmake args

simoncozens:
See #750.


RosaWagner:
Just tested it and it seems that it is working :)

PR #769, opened 2023-10-23
push: Get all files in a pr

### LINKED ISSUE #737: gen-push-lists only allows 100 files per PR
Paginating things in a GitHub GraphQL query is a nightmare. Currently the code we have does this:

https://github.com/googlefonts/gftools/blob/cc53fb57e939e0a1eeb256c5dcdb04caec5987cd/Lib/gftools/push.py#L83-L88

and then

https://github.com/googlefonts/gftools/blob/cc53fb57e939e0a1eeb256c5dcdb04caec5987cd/Lib/gftools/push.py#L106-L112

We correctly loop over multiple pages because we know that there may be more than 100 "items" in the project board, but we assume there are at most 100 files per PR; this isn't a good assumption in something like a big lang update or a catalogue-wide metadata update.

However (and this is where it gets difficult) to make that work you need to say "give me page 2 of the files on page 3 of the items", and you [can't use multiple pagination cursors in a GitHub GraphQL query](https://stackoverflow.com/questions/59192308/github-graphql-v4-api-nested-pagination-multiple-pagination-cursors-can-not-be) (See also [this issue](https://github.com/shurcooL/graphql/issues/69#issuecomment-843128806)).

So we may need to do this the inefficient way: grab all the items and their associated PRs first, paging through all the pages of items, and then do a query for *each* PR, paging through all the lists of files.

Urgh.

m4rc1e:
Currently, the GraphQL query will only fetch the first 100 files in a pr. This means that `gen-push-lists` won't add all the files to the `to_sandbox.txt` or `to_production.txt` files for large prs e.g https://github.com/google/fonts/pull/6624.

I thought about implementing a double cursor in the GraphQL query. However, it would've produced a lot of redundant data. My current approach is to fetch files using the rest api, only if the pr contains more than 100 files.

Fixes #737 

cc @simoncozens

### DIFF
--- a/Lib/gftools/gfgithub.py
+++ b/Lib/gftools/gfgithub.py
@@ -6,20 +6,20 @@ import urllib
 import time
 
 
-GITHUB_GRAPHQL_API = 'https://api.github.com/graphql'
-GITHUB_V3_REST_API = 'https://api.github.com'
+GITHUB_GRAPHQL_API = "https://api.github.com/graphql"
+GITHUB_V3_REST_API = "https://api.github.com"
 
 
 class GitHubClient:
     def __init__(self, repo_owner, repo_name):
-        if not 'GH_TOKEN' in os.environ:
+        if not "GH_TOKEN" in os.environ:
             raise Exception("GH_TOKEN environment variable not set")
-        self.gh_token = os.environ['GH_TOKEN']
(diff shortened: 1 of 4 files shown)


m4rc1e:
@simoncozens mind taking a look?

PR #774, opened 2023-10-24
rename-glyphs added. Rename glyphs in a font based on another font

m4rc1e:
@RosaWagner use this tool to rename the glyphs in the vtt ttfs.

### DIFF
--- a/Lib/gftools/scripts/rename_glyphs.py
+++ b/Lib/gftools/scripts/rename_glyphs.py
@@ -1,0 +1,30 @@
+"""
+Rename glyph names in font_1 with font_2.
+
+Usage:
+gftools rename-glyphs font_1.ttf font_2.ttf out font_1-renamed.ttf
+"""
+from fontTools.ttLib import TTFont
+from ufo2ft.postProcessor import PostProcessor, _reloadFont
+from argparse import ArgumentParser
+
+
+def main(args=None):
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("font_1")
+    parser.add_argument("font_2")
+    parser.add_argument("-o", "--out", default="out-renamed.ttf")
+    args = parser.parse_args(args)
+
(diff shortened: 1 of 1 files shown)


Lib/gftools/scripts/rename_glyphs.py
```diff
@@ -0,0 +1,30 @@
...
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("font_1")
+    parser.add_argument("font_2")
+    parser.add_argument("-o", "--out", default="out-renamed.ttf")
+    args = parser.parse_args(args)
+
+    font1 = TTFont(args.font_1)
+    font2 = TTFont(args.font_2)
+    
+    name_map = dict(zip(font1.getGlyphOrder(), font2.getGlyphOrder()))
```
simoncozens, line 22: You'll get an ugly error here if the fonts don't have the same number of glyphs, which of course they should.

Lib/gftools/scripts/rename_glyphs.py
```diff
@@ -0,0 +1,30 @@
...
+
+    font1 = _reloadFont(font1)
```
simoncozens, line 24: Is this really needed? It makes sense in the context of ufo2ft, but I'm not sure about here. Although I can see perhaps there's a need to ensure the font is fully decompiled before you start messing with glyph names (since everything in fontTools is referenced by name, not ID).

Either remove if it's not needed, or add a comment about why it is, because currently it's not obvious. (I just loaded the font and now I have to reload it?)

simoncozens (changes requested):

m4rc1e:
@RosaWagner I'll just check this works for your use case before I merge.

m4rc1e:
@RosaWagner it's working fine for me. It'll also rename the glyphs in the VTT TSI tables.

RosaWagner:
What I tested works

PR #778, opened 2023-10-26
Update to match internal copy of proto

rsheeter:
This fixes the parsing of Pushster noted in https://github.com/google/fonts/issues/6666


m4rc1e:
Thank you very much!

Issue #776, opened 2023-10-26
builder: Include font build dependencies in meta table


m4rc1e:
I'm contemplating storing the build dependencies in every font's "meta" table. We have a situation where many designers haven't pinned the dependencies so we don't truly know what was used. By genning this table at build time, we can be certain of the dependencies.

I've just checked pyftsubset and it'll drop this table by default, which seems like something we'd want to do for this use case. I'd love the fonts in the google/fonts repo to have this table. I don't mind if the fonts on [fonts.google.com](http://fonts.google.com/) don't have it.

I've talked about this idea with Cosimo but I'm interested in hearing from everyone else.

This idea has already been discussed in 2017, https://github.com/googlefonts/fontmake/issues/251.

In the next post, I'll workout how big this table will be using Cosimo's dependency lister, https://gist.github.com/anthrotype/531a425c8a0ba5ee975bc2ec8add7b82

m4rc1e:
Using Cosimo's script, we get the following sizes:

gftools: 2kb, 1kb (zipped)
fontmake: 369b, 438b (zipped)

I'll see what size I can get the gftools deps down if I remove the dependencies which are not used for generating fonts.

m4rc1e:
Actually, Idk if it's worth bothering trying to compress gftools more than just zipping it since the fonts we serve won't have this data. Are we able to spare 1kb?

anthrotype:
of course you want _either_ gftools or fontmake, not both, since fontmake is a dependency of gftools, you probably only want gftools' deps.

> the dependencies which are not used for generating fonts.

you could place things in extras..  maybe the gftools builder should be in its own distribution..

> Idk if it's worth bothering trying to compress

only if you care about retaining the data for shipping fonts, which I think we do not want

m4rc1e:
> of course you want either gftools or fontmake, not both, since fontmake is a dependency of gftools, you probably only want gftools' deps.

The size results I've posted were done by running your gist on gftools and fontmake separately. the gftools dependencies will list fontmake. Personally, I'd prefer if we use the gftools deps since it includes vttLib, ttfautohinter etc so it'll make remaking the fonts much easier,

> Idk if it's worth bothering trying to compress

>> only if you care about retaining the data for shipping fonts, which I think we do not want

Exactly so I'm tempted to leave it as is. Shall I go ahead and open a draft pr with a sketch?

PR #779, opened 2023-10-26
builder: wip add build dependencies to Debg table

m4rc1e:
Implements #776. I chatted to Cosimo after opening the issue and we both agreed that storing the dependencies in the "Debg" table is a better idea.


m4rc1e:
I guess the next piece we want is the ability to output a requirements.txt out from a font.

Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,163 @@
+"""
+Similar to pip freeze, but only listing the installed dependencies for the selected package.
+
+E.g.:
+    $ python freeze_requirements.py fonttools[ufo]
```
anthrotype, line 5: it's no longer named "freeze_requirements.py" I believe

Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,163 @@
...
+    fonttools==4.24.4
+    fs==2.4.13
+    pytz==2021.1
+    six==1.16.0
+
+Credit to Cosimo Lupo (anthrotype) for starting this as a gist:
+https://gist.github.com/anthrotype/531a425c8a0ba5ee975bc2ec8add7b82
+"""
+
+__requires__ = ["packaging", "importlib_metadata; python_version < '3.8'"]
```
anthrotype, line 18: you don't need this line, I added back then because I was fond of this [pip-run](https://github.com/jaraco/pip-run) tool but here it serves no purpose since gftools deps are declared in its setup.py

Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,163 @@
...
+    Dict,
+    Set,
+    Tuple,
+)
+from fontTools.ttLib import TTFont, newTable
+
+try:
+    import importlib.metadata as importlib_metadata
+except ImportError:
+    import importlib_metadata  # type: ignore
```
anthrotype, line 35: I think you no longer need this importlib_metadata backport since you should be already only supporting python_requires=">=3.8" in setup.py, if I am not mistaken

Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,163 @@
...
+    ttfont["Debg"] = debg
+
+
+def main():
+    import sys
+
+    try:
+        req_spec = sys.argv[1]
+    except IndexError:
+        sys.exit("usage: freeze_requirements.py PACKAGE_NAME")
```
anthrotype, line 151: change the script name

Lib/gftools/scripts/gen_debg_deps.py
```diff
@@ -0,0 +1,34 @@
...
+from argparse import ArgumentParser
+from gftools.builder.dependencies import requirements_to_debg_table
+from fontTools.ttLib import TTFont
+
+
+def main(args=None):
+    parser = ArgumentParser(description=__doc__)
+    parser.add_argument("font", help="Path to font")
+    parser.add_argument("--inplace", help="Overwrite font", action="store_true")
+    parser.add_argument("-o", "--out", help="Output path for new font")
```
anthrotype, line 19: this should be a "mutually exclusive group" of arguments:
https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group

simoncozens:
Could we please split this into two branches - one to add the utility, the other to update the builder? (Since the builder may be about to dramatically change...)

m4rc1e:
Closing since #780 has been merged. We'll integrate it into Builder2 next week.

PR #780, opened 2023-10-27
font-dependencies: tool and module added

m4rc1e:
Add font dependencies to a font's Debg table or dump requirements from a font file.

Splits #779.

### DIFF
--- a/Lib/gftools/builder/dependencies.py
+++ b/Lib/gftools/builder/dependencies.py
@@ -130,7 +130,7 @@ def format_requirements(requirements: Iterable[Tuple[str, str]]) -> str:
     return "".join(f"{name}=={version}\n" for name, version in sorted(requirements))
 
 
-def build_font_requirements(ttfont: TTFont, dist_name: str):
+def write_font_requirements(ttfont: TTFont, dist_name: str):
     if "Debg" in ttfont:
         debg = ttfont["Debg"]
     else:
@@ -143,7 +143,7 @@ def build_font_requirements(ttfont: TTFont, dist_name: str):
     ttfont["Debg"] = debg
 
 
-def dump_font_requirements(ttfont: TTFont):
+def read_font_requirements(ttfont: TTFont):
     debg_table = ttfont["Debg"].data
     return "\n".join(debg_table[GFTOOLS_DEPENDENCIES_KEY])
 
--- a/Lib/gftools/builder/dependencies.py
+++ b/Lib/gftools/builder/dependencies.py
@@ -1,0 +1,169 @@
+"""
+Similar to pip freeze, but only listing the installed dependencies for the selected package.
+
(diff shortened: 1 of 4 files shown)


Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,173 @@
...
+    ).splitlines()
+    ttfont["Debg"] = debg
+
+
+def dump_font_requirements(ttfont: TTFont):
+    assert "Debg" in ttfont, "No Debg table found!"
+    debg_table = ttfont["Debg"].data
+    assert (
+        GFTOOLS_DEPENDENCIES_KEY in debg_table
+    ), f"Missing '{GFTOOLS_DEPENDENCIES_KEY}' in Debg table"
```
anthrotype, line 151: the original KeyError would be better than an AssertionError. Assertions are supposed to never trigger in normal circumstances, whereas here one may well attempt to dump deps from a font that has none and an exception is the right approach in this case. If you don't like KeyError and want to customize the error message you can reraise it as whatever other exception you prefer

Lib/gftools/builder/dependencies.py
```diff
@@ -0,0 +1,173 @@
...
+        debg.data = {}
+    requirements = get_installed_requirements(dist_name)
+    debg.data["com.github.googlefonts.gftools.deps"] = format_requirements(
+        requirements
+    ).splitlines()
+    ttfont["Debg"] = debg
+
+
+def dump_font_requirements(ttfont: TTFont):
+    assert "Debg" in ttfont, "No Debg table found!"
```
anthrotype, line 147: same here

m4rc1e: I'm going to raise my own errors since the tools will be used by font designers and not just devs.

Lib/gftools/scripts/font_dependencies.py
```diff
@@ -0,0 +1,67 @@
+"""
+gftools font-dependencies
+
+Add the dependencies used to generate the font to the font's Debg table or
+dump a font's dependencies in order to make a pip requirements file.
+
+Usage:
+
+Build font dependencies:
+gftools font-dependencies build font.ttf -o font-deps.ttf
```
anthrotype, line 10: (bikeshedding) I think "store" reads better than "build" to describe the action this sub-command does. It's not really building anything, just retrieving and storing some piece of data in the font.
Or maybe 'read' and 'write' would also be appropriate.

Lib/gftools/builder/dependencies.py
m4rc1e: maybe it makes more sense for the tool to check the existence of the key and the table and raise a user friendly error there?

anthrotype (approved):

Issue #782, opened 2023-11-01
[builder] Temp files removed even when "cleanUp: False" set in the project.yaml file


aaronbell:
Since updating my python libraries to the point where the ninja builder is being used, I noticed that temp files are being deleted (specifically a master_ufo folder) despite having set `cleanUp: False` in the project.yaml file.

n8willis:
I don't mean to be a plusOneGuy here, but is there any chance of having this fix merged in?

It severely impedes the troubleshooting of build actions to not have access to the autogenerated files.

m4rc1e:
SGTM. I'll self assign it.

aaronbell:
Raising this again. The fix is really simple (a single `if` statement), someone just has to sign off on it.

PR #783, opened 2023-11-01
Update _ninja.py

aaronbell:
Making the Ninja script aware if "cleanUp" has been set in project.yaml. Otherwise Ninja will *always* delete all the temp files, even if that is undesirable (such as when running two separate project.yaml files that point to the same master_ufo directory). 

#782 - Resolves this issue

### DIFF
--- a/Lib/gftools/builder/_ninja.py
+++ b/Lib/gftools/builder/_ninja.py
@@ -63,19 +63,20 @@ class NinjaBuilder(GFBuilder):
                 os.remove(temporary)
 
         # Clean up temp build files
-        search_directory = os.getcwd()
-        target_names = ["build.ninja", ".ninja_log", "instance_ufo", "master_ufo"]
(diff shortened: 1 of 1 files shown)


aaronbell:
Here again asking if this can be merged since, yet again, I'm encountering the bug that this will fix.

simoncozens:
I understand the frustration but we are on the point of merging a rewrite to the builder (hopefully this week!), so fixes to the old one are superfluous.

aaronbell:
That's all well and good. But it has been 4 months with a simple, one-line fix sitting in a PR during which time there have been numerous releases and updates, including updates to _ninja.py. I don't see why this couldn't have been scooped up at some point rather than making folks encounter the same bug over and over.

m4rc1e:
Tests are failing since the runs have been deleted. It passed locally so I've merged the pr,

@aaronbell apologies and will cut you a release now.

aaronbell:
@m4rc1e Thanks Marc

PR #790, opened 2023-11-16
qa: add interpolation checks

m4rc1e:
Will merge once Behdad has finished and a new fontTools release is cut.


behdad:
I'm done for now.

PR #800, opened 2023-12-14
qa: return errcode 1 if fontbakery reports a fatal error

m4rc1e:
Will merge once a new fontbakery release is cut. PR implementation will simply log a fatal error and stop all other checks from running.

cc @simoncozens @felipesanches

### DIFF
--- a/Lib/gftools/qa.py
+++ b/Lib/gftools/qa.py
@@ -5,6 +5,7 @@ import traceback
 
 from gftools.gfgithub import GitHubClient
 from gftools.utils import mkdir
+import sys
 
 try:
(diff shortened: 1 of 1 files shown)


m4rc1e:
Ah, just realised this implementation won't post the fb report as a comment since it'll quit before this happens. Would it make more sense if FB posts the report as a comment and then raises the error code?

simoncozens:
LGTM. Please remember that `.ci/run.py` in google/fonts will also need `check=True` on its `subprocess.run` calls to propagate the error code.

felipesanches:
thanks! ;-)

PR #816, opened 2024-01-26
qa: display log message instead of raising an error if fontbakery report cannot be posted.

m4rc1e:
gftools qa can post fontbakery reports back as issue comments. This functionality isn't possible in the google/fonts ci if the pull request has come from a forked repo e.g https://github.com/google/fonts/pull/7136. Github Actions doesn't allow forked repos to access secrets since the user could just steal them. Instead of raising an error, I've decided to add a warn log message instead with the full traceback.

@simoncozens thoughts on this?


simoncozens:
SGTM!

PR #822, opened 2024-02-05
Replace glyphsets.codepoints with gfsubsets

simoncozens:
- Remove dead scripts
- Dead code
- Use gfsubsets instead of glyphsets.codepoints
- Require gfsubsets
- Opportunistic black


pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
   'absl-py',
   'glyphsLib',
   'gflanguages>=0.4.0',
+  'gfsubsets>=2024.02.05',
```
m4rc1e, line 35: is this a typo, or do dates actually work?

m4rc1e (commented):
LGTM if the date works.

pyproject.toml
simoncozens: Yeah, they work fine: major version 2024, minor version 2, patch version 5.

PR #831, opened 2024-02-14
Glyphsets API change

yanone:
This is using the new API of the new glyphsets database.

However, since the code mentions "subsets" in several instances, please make sure whether the gftools code really wants glyphsets or subsets here. I have a feeling that it wants subsets and this code is really totally out of place.

### DIFF
--- a/Lib/gftools/subsetmerger.py
+++ b/Lib/gftools/subsetmerger.py
@@ -12,7 +12,7 @@ import ufoLib2
 import yaml
 from fontmake.font_project import FontProject
 from fontTools.designspaceLib import DesignSpaceDocument
-from glyphsets import GFGlyphData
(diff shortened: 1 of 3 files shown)


simoncozens:
Sorry, the terminology of this whole program is a pain - it's about merging a subset of glyphs from a UFO into another UFO, but sometimes we choose those glyphs based on a range of Unicode codepoints. And sometimes we want to specify those Unicode codepoints based on a *glyphset* (i.e. "Merge the codepoints covered by GF Latin Core from Noto Sans into this UFO").

So the section of the code in this PR should refer to GF glyphsets, not (subsetter) subsets.

yanone:
I see. I've updated the glyphsets dependency in line with the API change, and believe this is ready to go now.

PR #832, opened 2024-02-15
Format HTML files

yanone:
There's a script called `gftools format-html`, and it's also auto-formatting all incoming html files in the packaging process.


m4rc1e:
Thanks! I'm going to merge since I also need this for push stuff. We can always improve it later.

PR #835, opened 2024-02-21
add-font: fix min_pct_ext default val

m4rc1e:
@emmamarichal pointed out that `add-font` will add `greek-ext` and `cyrillic-ext` subsets to greek and cyrillic families that don't contain the extended glyphs. In nam-files, we have, https://github.com/googlefonts/nam-files/blob/main/Lib/gfsubsets/__init__.py#L203-L204:

```PYTHON
        if 100.0 * len(overlap) / len(subset_cps) > target_pct:
            results.append((subset, len(overlap), len(subset_cps)))
```

We're multiplying the result by 100 so we're expecting an integer value, not a float which we've currently got. By changing this value to an int, it also matches `--min_pct` int.


simoncozens:
This makes more sense, the help text says "what percentage" and I always thought it was weird that it was 0.01. 0.01% is very little...

simoncozens (approved):

PR #837, opened 2024-02-22
builder._ninja: fix vf family name

m4rc1e:
For [Danfo](https://github.com/emmamarichal/danfo), the gftools builder is outputting the VF with the filename Danfo-Regular[ELSH].ttf. This PR will instead use the family name from the default source in the designspace, instead of the designspace filename.

cc @emmamarichal @simoncozens


m4rc1e:
@simoncozens  `.replace(" ", "")`?

m4rc1e:
Got smacked down by the tests. I'll think about this a bit more.

Issue #838, opened 2024-02-26
Undocumented `autohintOTF` fails many builds


bramstein:
A lot of our builds are failing with the new gftools version because it tries to run psautohint on the OTF files and many of our source files are not set up for this.

For now we can disable this behaviour by setting `autohintOTF` to false, but it would be good to review why this is now default and perhaps document it.

simoncozens:
You're right, this should be disabled by default.

It was default because I thought (a) auto hinting OTFs was probably a good idea (and someone asked for it), and crucially (b) because I thought `psautohint` was similar to `ttfautohint` in that you can just run it and it hints stuff automatically. I didn't realise it needed set up. I'll turn it off.

PR #839, opened 2024-02-27
Turn off PS autohinting by default

simoncozens:
See #838.


m4rc1e:
Thanks.

@bramstein would you like me to cut a new release for you?

PR #857, opened 2024-03-18
Automatic cleanups

simoncozens:
Trying to keep the code tidy with some automatic cleanups: `black`/`isort`/`ruff --fix` on everything.


simoncozens:
This needs doing again since we changed much stuff.

PR #860, opened 2024-03-18
Enforce code kwalitee

simoncozens:
("Kwalitee" is an old Perl joke -it's Michael Schwern's name for code quality that is automatically measurable; it's not actually a reflection of the *quality* of the code at all, but it's better than nothing.)

This adds black and pylint checks with a reasonable subset of pylint errors that we really should fix to get us going.


simoncozens:
(Of course tests will fail horribly until #857 is merged, and after that they will only fail a little bit.)

PR #883, opened 2024-03-22
Improve HTML formatter & activate it in new packager

yanone:


Lib/gftools/packager.py
```diff
@@ -302,6 +302,12 @@ def package_family(
                 os.remove(fp)
         shutil.copytree(tmp_dir, family_path, dirs_exist_ok=True)
         save_metadata(family_path / "METADATA.pb", metadata)
+        # Format HTML
+        if os.path.exists(os.path.join(family_path, "DESCRIPTION.en_us.html")):
+            with open(os.path.join(family_path, "DESCRIPTION.en_us.html"), "r") as fin:
+                description = format_html(fin.read())
+            with open(os.path.join(family_path, "DESCRIPTION.en_us.html"), "w") as fout:
+                fout.write(description)
```
m4rc1e, line 310: `family_path` is a `pathlib.Path` object so we don't need to use `os.path`. I don't think we need to check if a description exists because `save_metadata` will add one if one does't exist.

```suggestion
        desc_file = family_path / "DESCRIPTION.en_us.html"
        with open(desc_file, encoding="utf-8") as fin:
            description = format_html(fin.read())
        with open(desc_file, "w", encoding="utf-8") as fout:
            fout.write(description)
```

Lib/gftools/utils.py
```diff
@@ -278,6 +278,32 @@ def _html_custom_formatter(string):
...
+                or (
+                    re.search(r"\W\w{1,2}[\.|\?|!]$", this_line)
+                    and this_line[-2] == this_line[-2].upper()
+                    and next_line[1] == next_line[1].lower()
+                )  # H.R. Giger
+            ):
+                strings[i] = strings[i] + strings[i + 1]
+                strings[i + 1] = ""
+    # Join back together
+    string = "\n".join(strings)
```
m4rc1e, line 304: We could use a range to make this part cleaner

```suggestion
    for i in range(1, len(strings):
        this_line = strings[i-1]
        next_line = strings[i]
        if next_line = "":
            continue
        if (
            re.search(r"i\.?e\.$", this_line)  # ie.
            or re.search(r"e\.?g\.$", this_line)  # eg.
            or (
                re.search(r"etc[\.|\?|!]$", this_line)
                and next_line[1] == next_line[1].lower()
            )  # etc.
            or (
                re.search(r"\W\w{1,2}[\.|\?|!]$", this_line)
                and this_line[-2] == this_line[-2].upper()
                and next_line[1] == next_line[1].lower()
            )  # H.R. Giger
        ):
            strings[i-1] = strings[i-1] + strings[i]
            strings[i] = ""
    # Join back together
    string = "\n".join(strings)
```

yanone: As you prefer Marc. Feel free to rewrite it, as the test will ensure the correct formatting.

m4rc1e: Let me test this first.

m4rc1e (approved):
Thank you very much for the testcases.

PR #888, opened 2024-03-26
Fix glyphData (again)

khaledhosny:
Apparently I’m the only one who uses this and it keeps breaking, so this time make sure it gets tested.

### DIFF
--- a/Lib/gftools/builder/recipeproviders/googlefonts.py
+++ b/Lib/gftools/builder/recipeproviders/googlefonts.py
@@ -155,13 +155,13 @@ class GFBuilder(RecipeProviderBase):
                 args += " --no-check-compatibility"
             if self.config.get("extraVariableFontmakeArgs") is not None:
                 args += " " + str(self.config["extraVariableFontmakeArgs"])
+            if source.is_glyphs:
+                for gd in self.config.get("glyphData", []):
+                    args += " --glyph-data " + gd
         else:
             if self.config.get("extraStaticFontmakeArgs") is not None:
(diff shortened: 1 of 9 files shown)


khaledhosny:
Any idea why all these unrelated tests are failing?

Lib/gftools/builder/recipeproviders/googlefonts.py
```diff
@@ -221,10 +221,6 @@ def build_a_static(self, source, instance, output):
         target = os.path.join(outdir, f"{instancebase}.{output}")
 
         static_args = ""
-        if source.is_glyphs:
-            for gd in self.config.get("glyphData", []):
-                static_args += " --glyph-data " + gd
```
simoncozens, line 226: Why do we want to remove this for statics?

m4rc1e:
> Any idea why all these unrelated tests are failing?

@khaledhosny one of our test servers is currrently experiencing issues. We'll ignore this for the time being.

simoncozens:
The refactoring I did in #906 will clobber this.

khaledhosny:
> The refactoring I did in #906 will clobber this.

I rebased it on current master and kept only the test. Building variable fonts is fixed, but building statics is still broken.

khaledhosny:
> > The refactoring I did in #906 will clobber this.
> 
> I rebased it on current master and kept only the test. Building variable fonts is fixed, but building statics is still broken.

This should be fixed now. The tests now actually check that glyphData is being applied, not only that the build passes.

khaledhosny:
The test failures does not seem to be related to this PR.

Issue #896, opened 2024-03-28
packager: inherit labels from an issue


m4rc1e:
cc @yanone

yanone:
Also: link issue with PR.
And in case it's not already happening: Add PR to Traffic Jam project.

PR #909, opened 2024-04-05
Packager fixes

### LINKED ISSUE #665: [packager] Link PR with issue
Just typing down a thought that would greatly enhance the packaging workflow. I've had it ever since we transitioned to fully leveraging Github project management, but it's becoming more important now that we're using the new project boards which don't allow linking PRs with issues directly from the board, which the old projects allowed:

When submitting a font with the packager, a Github issue ID should be provided _mandatorily_ as soon as `--pr` is used, and be omittable only with `--noissue`.

Then the PR can be automatically linked with the issue and all other measures taken to label/assign the PR correctly following whatever data can be derived from the issue, and also the issue status changed.

`gftools packager whateverfont.yaml ./fonts --pr -i 3667` or
`gftools packager whateverfont.yaml ./fonts --pr --noissue`

I think it should be the `packager` command where the linking takes place.

@m4rc1e @RosaWagner

### LINKED ISSUE #880: packager: better error messages if files not found
Currently, you'll just get a github 404 if a file isn't found. It may be useful to suggest the closest file.

m4rc1e:
- Better error message fixes #880 
- Add issue link fixes #665 
- Fix pr labelling #896

### DIFF
--- a/Lib/gftools/packager.py
+++ b/Lib/gftools/packager.py
@@ -549,8 +549,10 @@ def make_package(
         log.fatal(
             f"'{metadata.name}' does not have family tagging data! "
             "Please complete the following form, "
-            "https://forms.gle/jcp3nDv63LaV1rxH6. This is a hard requirement "
-            "set by Google Fonts management."
+            "https://forms.gle/jcp3nDv63LaV1rxH6. Once tags have been added, "
+            "you may need to wait around five minutes in order for the tags "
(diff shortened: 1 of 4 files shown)


yanone:
Still to do: If issue # is given, copy all labels from the issue to the PR (such as "Small Fix", "Bug", etc)

PR #916, opened 2024-04-10
[packager] Improve error message when branch could not be created

simoncozens:
If the git branch command failed for any reason, you would get the error `'NoneType' object has no attribute 'name'` when the code tried to use the (None) `branch` variable. This gives you the output from git.


m4rc1e:
Looks good. However, I'm interested in understanding why it wasn't able to create the branch?

simoncozens:
I've worked out why this was failing for me: I have git configured to pull with rebase instead of merge, to avoid needless merges. But that only works if the working tree is clean. `_create_git_branch` does a `git pull --force` but we are guaranteed to have a dirty working tree as we've just been editing a METADATA.pb file in there. So the `git pull` fails.

PR #918, opened 2024-04-11
set-overlap-bits added

m4rc1e:
This script will automatically set the glyph overlap bits for glyphs that have overlapping contours. Atm, it will only work on ufos since AFAIK .glyphs files do not have the ability to set these bits yet.

I originally wrote this as a ufo2ft filter. However, on David Berlow families, it was agonisingly slow. The pain was similar to installing win 95 using floppy disks.

I may be cheeky and throw this into the gftools builder for fonts that have less than 10 masters.

### DIFF
--- a/Lib/gftools/scripts/set_overlap_bits.py
+++ b/Lib/gftools/scripts/set_overlap_bits.py
@@ -1,0 +1,57 @@
+import argparse
+from defcon import Font
+from pathops.operations import _draw
+from ufo2ft.filters.decomposeComponents import DecomposeComponentsFilter
+from ufo2ft.preProcessor import TTFPreProcessor
+import math
+from fontTools.designspaceLib import DesignSpaceDocument
+
+
+def set_overlap_bits(ufo):
+    # Skip setting bits if ufo
+    if any(g.lib.get("public.truetype.overlap") for g in ufo):
+        return
+    # Decompose components first because some component glyphs may have
+    # components that overlap each other
+    outline_glyphset = TTFPreProcessor(
+        ufo, filters=[DecomposeComponentsFilter()]
+    ).process()
+
+    overlaps = set()
+    for glyph in outline_glyphset.values():
+        paths = _draw(glyph)
+        area = paths.area
+        # rm overlaps
+        paths.simplify()
+        simplified_area = paths.area
+        if math.ceil(area) != math.ceil(simplified_area):
+            ufo[glyph.name].lib["public.truetype.overlap"] = True
+            overlaps.add(glyph.name)
+    return overlaps
+
+
+def main(args=None):
+    parser = argparse.ArgumentParser(description="Set the overlap bits of a ufo/ds")
+    parser.add_argument("input", help="Input UFO or Designspace file", nargs="+")
+    args = parser.parse_args(args)
+
+    ufos = []
+    for fp in args.input:
+        if fp.endswith(".ufo"):
+            ufos.append(Font(fp))
+        elif fp.endswith(".designspace"):
+            ds = DesignSpaceDocument.fromfile(fp)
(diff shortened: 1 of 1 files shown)


Lib/gftools/scripts/set_overlap_bits.py
```diff
@@ -0,0 +1,60 @@
...
+    for glyph in outline_glyphset.values():
+        skia_path = Path()
+        pen = skia_path.getPen()
+        for contour in glyph:
+            contour.draw(pen)
+        area = skia_path.area
+        # rm overlaps
+        skia_path.simplify()
+        simplified_area = skia_path.area
+        if not math.isclose(area, simplified_area, abs_tol=0.1):
```
anthrotype, line 30: it's unusual to set abs_tol so much larger than rel_tol (default rel_tol is 1e-09), it'd take precedence most of the time, whereas in fact it's supposed to help with comparisons close 0 (where rel_tol tends to be too small), see https://docs.python.org/3/library/math.html#math.isclose
If you only care about comparing against an absolute value (and not do any relative comparisons) just do abs(a-b) <= abs_tol and you don't need to bother using isclose.

But actually, perhaps it would make more sense to use a tolernce that is relative to the font's unitsPerEm?
I don't know, something like `font.info.unitsPerEm / 10_000` (you'd still get 0.1 for fonts with 1000 UPEM, but if would increase/decrease accordingly)

Lib/gftools/scripts/set_overlap_bits.py
```diff
@@ -0,0 +1,60 @@
...
+
+    overlaps = set()
+    for glyph in outline_glyphset.values():
+        skia_path = Path()
+        pen = skia_path.getPen()
+        for contour in glyph:
+            contour.draw(pen)
+        area = skia_path.area
+        # rm overlaps
+        skia_path.simplify()
```
anthrotype, line 28: this can fail sometimes (pathops has some tricky unfixable bugs), I think it'd be better to default to setting the flag for all glyphs _except_ those whose area does _not_ change after simplifying

Lib/gftools/scripts/set_overlap_bits.py
```diff
@@ -0,0 +1,60 @@
...
+    args = parser.parse_args(args)
+
+    ufos = []
+    for fp in args.input:
+        if fp.endswith(".ufo"):
+            ufos.append(Font(fp))
+        elif fp.endswith(".designspace"):
+            ds = DesignSpaceDocument.fromfile(fp)
+            for src in ds.sources:
+                ufos.append(Font(src.path))
```
anthrotype, line 48: doing this for each master and then for each individual glyph is not sufficient; you want to set the flag on _all_ the masters (or if you like only on the default master, which is the one that matters and which will contribute the actual glyf table to the final VF) if _any_ of them needs overlap. If you only set on a non-default master, but not on the default one, then it will just be ignored and not present in the VF.

Lib/gftools/scripts/set_overlap_bits.py
```diff
@@ -0,0 +1,60 @@
...
+from pathops import Path
+from ufo2ft.filters.decomposeComponents import DecomposeComponentsFilter
+from ufo2ft.preProcessor import TTFPreProcessor
+import math
+from fontTools.designspaceLib import DesignSpaceDocument
+
+
+def set_overlap_bits(ufo):
+    # Skip setting bits if ufo
+    if any(g.lib.get("public.truetype.overlap") for g in ufo):
```
anthrotype, line 12: safe to skip if flags are present (might have been added manually), but perhaps you could also add a CLI flag to overwrite existing overlap flags, as they might have been mechanically generated and out of sync

PR #922, opened 2024-04-19
Keep glyph names when subsetting

simoncozens:
See https://github.com/TypeTogether/Playwrite/issues/38. Running hbsubset causes the post table to be upgraded to version 3.0, but we warn against using that.


simoncozens:
(Not my fail, I'll fix that in another PR.)

PR #927, opened 2024-04-25
[builder] Choose subsetter

simoncozens:
This allows the user to choose between hb-subset and pyftsubset for subsetting. (The operation name now feels a bit stupid but I feel like it's probably too late to change that now.)

To determine the subsetter to use, it looks for "subsetter" in the operation step, and if it's not there, looks in the `GFTOOLS_SUBSETTER` environment variable. If that's "python" you get `pyftsubset`; if it's "harfbuzz" you get `hb-subset`; if it's "auto" (or not provided), you get `hb-subset` if it's installed and `pyftsubset` otherwise.

### DIFF
--- a/Lib/gftools/builder/operations/hbsubset.py
+++ b/Lib/gftools/builder/operations/hbsubset.py
@@ -1,6 +1,36 @@
+import logging
+import os
+import shutil
+
 from gftools.builder.operations import OperationBase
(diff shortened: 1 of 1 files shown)


m4rc1e:
Thank you so much! Should fix the failing ci.

PR #929, opened 2024-04-25
[gftools-fix] Restructuring

simoncozens:
This started off as a simple PR (add a fix for the [hhea.caretSlopeRun](https://github.com/TypeTogether/Playwrite/issues/43#issuecomment-2076079076) issue caused by interpolating an MVAR table), but it's turned into a bit of rewrite:

* All fix functions are now expected to have a consistent interface: take a TTFont, return a tuple `changed, messages`. Typing has been added to ensure this.
* The "check if the current value of a field is what we expect, and if not, update it, emitting a message" pattern has been abstracted out into a function and used consistently.
* All fix functions check if they are appropriate (have all the tables they need etc.) and return `False, []` if not.

With this in place, `fix_font` becomes a matter of iterating over a list of fix functions.

Additionally, I added metadata to the functions linking them to a Fontbakery fix. Currently this metadata isn't used, but it would be helpful eventually to keep things automatically in sync, and to automatically apply fixes to failing FB checks.

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -460,6 +460,20 @@ def fix_italic_angle(ttFont) -> FixResult:
     return False, []
 
 
+def fix_hhea_caret_slope_run(ttFont: TTFont) -> FixResult:
+    if ttFont["post"].italicAngle == 0:
+        return False, []
+
+    expected_rise = ttFont["head"].unitsPerEm
+    expected_run = round(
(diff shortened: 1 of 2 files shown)


m4rc1e:
Thanks Simon. I'll take a look on Monday.

m4rc1e (approved):
Love it. Thank you very much.

PR #955, opened 2024-05-21
Add custom GFArgumentParser

m4rc1e:
@simoncozens I've been toying with the idea of having our own `ArgumentParser` class. By default, it'll have the `--show-tracebacks` and `--log-level` commands set to sensible defaults. Once fully implemented, I could replace all ArgumentParsers with our one quite easily.

Still WIP since this may be a footgun idea.


simoncozens:
Yes, like it!

PR #963, opened 2024-05-28
fontsetter flexibility

simoncozens:
This adds two new features to gftools-fontsetter:

* A value of the form `"|= 0x123"` will OR the current value of the field with the provided value.
* Multiple keys can be provided in the same configuration file. This allows for calling the same method (`name->setName`) multiple times. (Needed for making things italic.)


m4rc1e:
sgtm!

PR #964, opened 2024-05-28
Split Italic and Roman VFs in builder

### LINKED ISSUE #957: [builder] If font has ital axis, subspace into Roman and Italic
As per [GF Guide](https://googlefonts.github.io/gf-guide/variable.html#ital--slnt).

simoncozens:
Fixes #957.

We sometimes have source files containing an `ital` or `slnt` axis, but we require the onboarding of two separate fonts with Roman and Italic joined by style linking and STAT axis. This PR uses `varLib.instancer` to partially instance them along the relevant italic axis and then fixes them up accordingly.

Requires #963.

### DIFF
--- a/Lib/gftools/builder/recipeproviders/googlefonts.py
+++ b/Lib/gftools/builder/recipeproviders/googlefonts.py
@@ -3,10 +3,13 @@ import logging
 import os
 import re
 from tempfile import NamedTemporaryFile
+from typing import Optional, Tuple
 
+import babelfont
 import yaml
 from strictyaml import load, YAMLValidationError
 
+from gftools.builder.file import File
 from gftools.builder.recipeproviders import RecipeProviderBase
 from gftools.builder.schema import (
     GOOGLEFONTS_SCHEMA,
@@ -16,6 +19,8 @@ from gftools.builder.schema import (
 from gftools.utils import open_ufo
 
 logger = logging.getLogger("GFBuilder")
+
+Italic = Optional[Tuple[str, float, float]]  # tag, regular value, italic value
 
 # Things not ported from old builder:
 #  - cleanup
@@ -50,6 +55,7 @@ DEFAULTS = {
     "useMutatorMath": False,
     "checkCompatibility": True,
     "overlaps": "booleanOperations",
+    "splitItalic": True,
 }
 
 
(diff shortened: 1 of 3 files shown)


arrowtype:
Would this accurate handle discrete (non-continuous) Italic axes, or is that a separate issue?

For example, I’d like to do the following:

``xml
    <axis tag="ital" name="Italic" values="0 1" default="0">
      <labels>
        <label uservalue="0" name="Roman" elidable="true" />
        <label uservalue="1" name="Italic" />
      </labels>
    </axis>
```

This currently requires two designspace files, which pre-splits the axis between Roman and Italic sub-families. Luckily, this approach does seem to work well and create the STAT tables I want. Still, it requires some repeated information, which FontMake alone doesn’t require. I’m not sure how challenging this would be to map to the recipe logic of the Builder, though. If it’s not feasible, the builder still offers enough advantages that I would stick with it over directly using FontMake, but it would be handy to understand if the above axis setup is planned to be supported, or not.

simoncozens:
It probably *would* handle that if you have a single VF which is subspacable. I'm not sure what fontmake would do with that kind of designspace right now. Another option is DS5 designspaces which allow you to generate multiple VFs from the same designspace, but we absolutely don't support those quite yet and I'm not sure what tooling generates them anyway.

arrowtype:
> Glyphs.app's Variable Font Export Settings ... I think this is mostly to be able to set different custom parameters on a VF export, but I'm not sure if these Glyphs.app VF Export settings would allow one to define a particular axis range.

@anthrotype having used this feature quite a bit recently, it is basically there to define the naming info, especially if the family name is different or suffixed (e.g. "Familyname Variable") from the general family name given to static exports. From what I can see, the VF export settings and custom parameters don’t yet give any options around defining subsetting axis ranges, unlike the  `<variable-fonts>` elements of a DSv5.

emmamarichal:
@simoncozens 

I have a font with a backslant, where I need the backslant, and I have to get something like this: Myfont[slnt,wght].ttf. 
Now the builder split the file in two, and I would like to know how to change the stat table or anything in the build process to avoid this?

simoncozens:
To keep a single file (turn off the splitting), set `splitItalic: false` in the config.yaml.

PR #978, opened 2024-06-07
Add subsets check to PR checklist

yanone:


simoncozens:
I understand the intention, but if this is not happening then there's a bug in `add-font` in gftools.

yanone:
Found just now that Fustat was also missing the Arabic subset. While that's surely a bug, I implore you to add this check to the list, because it hurts no one to spend that one second of attention to the subsets.

m4rc1e:
I'm merging this but we should work on a better way to determine what subsets are needed such as #982

PR #987, opened 2024-06-12
Support Git references in subset merger

### LINKED ISSUE #986: Support Git references in subset merger from repo
Instead of always grabbing the latest of the default branch

References (thanks Simon!):

https://github.com/googlefonts/gftools/blob/c14661a8fdbfee8293dda9f6ff2994c173672019/Lib/gftools/subsetmerger.py#L303-L314

https://github.com/googlefonts/gftools/blob/c14661a8fdbfee8293dda9f6ff2994c173672019/Lib/gftools/scripts/add_ds_subsets.py#L105-L112

I'll hopefully get to this this week

RickyDaMa:
Closes #986 

In YAML, you can now do:

```yml
- operation: addSubset
  directory: latin-sources
  subsets:
    - from:
        repo: googlefonts/some-font@latest # <- the @ bit is new!
        path: source/some-font/SomeFont.designspace
      force: true
      name: GF_Latin_Core
```

Which can take any git reference, or the special value "latest" which grabs the tag from the latest GitHub release

These are appropriately cached in separate folders for org/repo/ref

Support for this is has also been incorporated into `add-ds-subsets`, and I've sprinkled in some type hints in parts of the codebase I've visited

<details>
<summary>Old PR description (outdated)</summary>

In YAML, you can now do:

```yml
repo:
  slug: notofonts/latin-greek-cyrillic
  ref: e7f1736c5ad0dc2abfc4dcd49ebca50abf612b29
```

Where before repo could only take a string (being the repo slug). This is still supported with the old behaviour of assuming you want the latest of the branch

The cache path was previously just the repo slug, but is now the slug + ref, e.g. `notofonts/latin-greek-cyrillic/main`

I've added a corresponding `--git-ref` to `gftools-add-ds-subsets` to expose this feature to the CLI as well

Opening this PR early without tests etc. to make sure the approach & code style are acceptable - such as the smidge of type hinting I added

Let me know your thoughts!
</details>

### DIFF
--- a/Lib/gftools/gfgithub.py
+++ b/Lib/gftools/gfgithub.py
@@ -88,6 +88,9 @@ class GitHubClient:
         """https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release"""
         return self._get(self.rest_url("releases/latest"))
 
+    def get_latest_release_tag(self) -> str:
+        return self.get_latest_release()["tag_name"]
+
     def open_prs(self, pr_head: str, pr_base_branch: str) -> typing.List:
         return self._get(
             self.rest_url("pulls", state="open", head=pr_head, base=pr_base_branch)
--- a/Lib/gftools/gfgithub.py
+++ b/Lib/gftools/gfgithub.py
@@ -85,6 +85,7 @@ class GitHubClient:
         return response
 
     def get_latest_release(self):
+        """https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release"""
         return self._get(self.rest_url("releases/latest"))
 
     def open_prs(self, pr_head: str, pr_base_branch: str) -> typing.List:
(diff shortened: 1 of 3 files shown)


simoncozens:
This is good, but:

1) I'd rather "repo" be the repo slug and a new optional "ref" field be added. There's no need for an additional level of nesting here.
2) Type hints are good and I want more of them everywhere! But I don't think we need `from __future__ import annotations` unless you're doing self-referential typing ("Postponed Evaluation of Annotations"). Type hints are supported by default in all the Python versions we support.

simoncozens:
> Or we could do pip-like syntax and have repo support something like org/repo@git-ref

I like that.

> They're to be able to use lowercase type hints for collections in Python 3.8

Hmm. OK, fair enough!

simoncozens:
> shall I just hack something to make it work?

There's only one top-level folder, right? So I guess you can just glob for it.

m4rc1e:
Thanks for the force push. Let's see if I've fixed it.

simoncozens (approved):

PR #994, opened 2024-06-26
ci: do not run tests that need secrets from forks

### LINKED ISSUE #988: Fix CI pipelines for 3rd party contributors
It would be nice to get helpful feedback from CI as an external contributor instead of needing to know which tests are expected to fail and then trying to see if there are any other problems

Maybe some of these tests should just be skipped if being run outside of the googlefonts organisation?

https://github.com/googlefonts/gftools/pull/870#issuecomment-2009606566

m4rc1e:
Fixes #988

cc @RickyDaMa

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,6 +58,8 @@ jobs:
         if: github.event.pull_request.head.repo.fork == false
         run: |
           pytest tests/push/*.py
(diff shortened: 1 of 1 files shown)


m4rc1e:
No worries. I apologise about the delay to these fixes. I'm juggling many hats atm.

PR #1006, opened 2024-07-12
CJK subset restriction added to checklist

vv-monsalve:
We need to Warn about this. It will have a FATAL status


m4rc1e:
> We need to Warn about this. It will have a FATAL status

Already in place ;-) https://github.com/fonttools/fontbakery/pull/4787

PR #1007, opened 2024-07-15
fix: Wrong escape sequences

### LINKED ISSUE #1004: [Builder] sometimes fails at OTFautohint step because it tries to build `||` file
I’m not really sure what’s happening here, but I’ve run into the same issue intermittently, so I wanted to report it. It only happens occasionally, and tends to sort itself out with a re-run, so it’s nothing super serious. But, it is a little confusing, and could be a problem worth solving.

Basically, it seems like the GF Builder will occasionally create a file called simply `||`, then try to run it through the OTFautohint step. This doesn’t work, so it gives me the following error:

```
fontmake --output-path /tmp/tmpm2b7lj3i -o otf -u sources/ufo/instance_ufos/Familyname-Black.ufo.json --filter ... --filter FlattenComponentsFilter --filter DecomposeTransformedComponentsFilter
[24/31] autohintOTF
FAILED: /tmp/tmptz84cwqz 
/gftools/bin/python -m gftools.builder.jobrunner otfautohint  -o /tmp/tmptz84cwqz /tmp/tmpc31o27cq \|\| otfautohint  -o /tmp/tmptz84cwqz /tmp/tmpc31o27cq --no-zones-stems 

Command failed:
otfautohint -o /tmp/tmptz84cwqz /tmp/tmpc31o27cq || otfautohint -o /tmp/tmptz84cwqz /tmp/tmpc31o27cq --no-zones-stems

usage: otfautohint [-h] [-v]
                   [-g GLYPH_LIST | --glyphs-file PATH | -x GLYPH_LIST | --exclude-glyphs-file PATH]
                   [-m GLYPH_LIST | --overlaps-file PATH | --force-overlap | --force-no-overlap]
                   [--log PATH] [-p PROCESSES] [--version] [--traceback]
                   [-o PATH [PATH ...]] [-r PATH] [-b PATH] [-a] [-w] [-k]
                   [-c | --report-only] [-d] [--no-flex] [--no-hint-sub]
                   [--no-zones-stems] [--fontinfo-file PATH]
                   [--ignore-fontinfo]
                   [--print-list-fddict | --print-all-fddict] [--doc-fontinfo]
                   [--info] [--max-segments MAX_SEGMENTS] [--test]
                   [FONT ...]
otfautohint: error: argument -o/--output: /fonts/Familyname/|| is not a valid path to write to.
```

This has happened for various different fonts, and it generally works itself out if I simply re-run the build. I’m not sure what causes it, but because it works out without any changes from me, I don’t think it’s an issue in my fonts or my config files. As a total guess, maybe it’s something like a race condition, or a random glitch of temporary file naming, or something similar and slightly unpredictable. I mostly just wanted to record it in case others run into this, or in case there’s an obvious solution. 

I’m on `gftools 0.9.61`, and my current config file looks like this:

```
familyName: Familyname
buildSmallCap: false
buildStatic: true
buildVariable: false
buildWebfont: false
autohintOTF: true
autohintTTF: true
interpolate: true
outputDir: ./fonts
sources:
  - sources/ufo/Familyname.designspace
recipeProvider: googlefonts
recipe:
  fonts/variable/FamilynameVariable[wdth].ttf:
    - source: sources/ufo/Familyname.designspace
    - args: --filter ...  --filter FlattenComponentsFilter --filter DecomposeTransformedComponentsFilter
      operation: buildVariable
    - args: "fix-nonhinting $in $out"
      exe: "gftools"
      operation: exec
    - args: "../scripts/rename-fonts.py $in --new_name 'Familyname Variable' && cp $in $out"
      exe: "python3"
      operation: exec
```

aragon999:
Maybe fixes #1004

### DIFF
--- a/Lib/gftools/builder/operations/autohintOTF.py
+++ b/Lib/gftools/builder/operations/autohintOTF.py
@@ -3,4 +3,4 @@ from gftools.builder.operations import OperationBase
 
 class AutohintOTF(OperationBase):
     description = "Run otfautohint"
-    rule = "otfautohint $args -o $out $in \|\| otfautohint $args -o $out $in --no-zones-stems"
+    rule = "otfautohint $args -o $out $in \\|\\| otfautohint $args -o $out $in --no-zones-stems"
--- a/Lib/gftools/push/items.py
+++ b/Lib/gftools/push/items.py
@@ -212,7 +212,7 @@ def parse_html(string: str):
     if not string:
         return None
     text = BeautifulSoup(string, features="lxml").text
-    return re.sub("\s+", " ", text).strip()
+    return re.sub(r"\s+", " ", text).strip()
 
 
 @dataclass


Lib/gftools/builder/operations/autohintOTF.py
```diff
@@ -3,4 +3,4 @@
 
 class AutohintOTF(OperationBase):
     description = "Run otfautohint"
-    rule = "otfautohint $args -o $out $in \|\| otfautohint $args -o $out $in --no-zones-stems"
+    rule = "otfautohint $args -o $out $in \\|\\| otfautohint $args -o $out $in --no-zones-stems"
```
m4rc1e, line 6: @arrowtype's traceback shows,

```
otfautohint: error: argument -o/--output: /fonts/Familyname/|| is not a valid path to write to.
```

The or operator looks well escaped. The issue is that the or operator is attached to the filename. Perhaps quoting the $out and $in args will work e.g

```
'otfautohint $args -o "$out" "$in" \|\| otfautohint $args -o "$out" "$in" --no-zones-stems'
```

aragon999: You are right, probably the `$in` and `$out` arguments should be escaped with quotes. I will adjust that. The thing I encountered was actually a Python warning:
image

arrowtype:
@m4rc1e Sorry it took so long for me to see your request to try this. I don’t face this issue _all_ the time, and often, it just requires me re-running the build one or two times, and then it works. So, that’s odd, but seems to work.

However, I did test this branch, and it worked the first time, and a second time. The code changes also appear sensible enough to me, FWIW.

aragon999:
I fixed the merge conflict and would be happy if this gets merged, as this also fixes a `SyntaxWarning` since Python 3.12:
```
/usr/lib/python3.13/site-packages/gftools/push/items.py:220: SyntaxWarning: invalid escape sequence '\s'
  return re.sub("\s+", " ", text).strip()
```

As the PR did: https://github.com/googlefonts/gftools/pull/1073

Or should I split this PR up?

PR #1013, opened 2024-08-05
font-tags: turn into local webapp

m4rc1e:


rsheeter:
I don't think we want users to run python, just give them a url. github pages or even codepen (poc https://codepen.io/rs42/pen/mdZwdZM) seems likely to be sufficient unless there is some capability that requires python?

Lib/gftools/tag-templates/index.html
```diff
@@ -0,0 +1,183 @@
...
+      })
+      .then(response => response.json())
+      .then(data => {
+        console.log('Success:', data);
+      })
+      .catch((error) => {
+        console.error('Error:', error);
+      });
+    },
+    loadCSV() {
```
davelab6, line 124: Can this be redone to use the Github API?

davelab6 (approved):

davelab6:
> unless there is some capability that requires python?

Well, the workflow Marc actually has is editing the file locally, so, in the httpd() call he has,

>                "Once you're done editing, remember to commit the changes in "
>                "your google/fonts repo and open a pr."

But all the py code is doing is loading the CSV and serving it via a httpd; this is useful for a local workflow since then the file is edited and can be directly committed. 

For a github pages version, I think "all" Marc has to do is use the Github API (using something like https://github.com/elwayman02/ember-data-github) to load and commit the tags csv file directly. 

Both workflows are worth supporting, imo

rsheeter:
I would just let the user download either an updated file or a diff to use for a PR. That simplifies things and is a major improvement over the spreadsheet workflow. Most of the benefit at very low cost, near-0 maintenance because there are no dependencies or scary operations.

m4rc1e:
SGTM. I'll drop the server and just host it on the google/fonts gh-pages.

davelab6:
MVP :)

rsheeter:
> MVP :)

+1. Launch, begin to capture value, and iterate.

m4rc1e:
Closing since this has moved to google/fonts, https://google.github.io/fonts/tags.html

PR #1030, opened 2024-09-20
Cut instances in the Noto recipe provider

simoncozens:
Now I've done various bits of optimisation, a good chunk of the time of a Noto build is caught up in instantiating UFOs in order to build static TTFs. I'd like to move towards cutting instances from VFs (that's what the front-end does, so it should be acceptable)...

However, this is currently draft because there are some differences between the cut instances and the UFO-instantiated builds, listed below. So we probably need a `gftools-cut-instance` which fixes up the problems, and use that instead of purely calling `hb-subset`.

* `winAscent` and `winDescent` are not updated, neither with `fontTools varLib.instancer` or `hb-subset`.
* `hb-subset` does not correct the macStyle and fsSelection bits for bold weights, leading to problems with style linking.  (fontTools OK here)
* I got multiple `.notdef` glyphs with `hb-subset`. (fontTools OK here)
* `hb-subset` does not update the name table (fontTools does).

### DIFF
--- a/Lib/gftools/builder/operations/hbsubset.py
+++ b/Lib/gftools/builder/operations/hbsubset.py
@@ -1,3 +1,4 @@
+from functools import cached_property
 import logging
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


yanone:
FWIW, I think the win metrics are not supposed to get updated. We have a FB check that complains if members of a family don't share the same win metrics.

simoncozens:
That's GF policy, yes. But it's not OT spec. If the masters *do* have varying win metrics, then in general a static instance cutter should update them.

Issue #1014, opened 2024-08-05
It appears "cleanup: False" no longer works


aaronbell:
If I do not want ninja to clean up the build files, for whatever reason, it would be helpful to set `cleanup: False` in the builder config. However, it does not appear to do anything anymore as it is not mentioned anywhere else in the builder code. I think it would be helpful to be able to disable cleanup.

davelab6:
@m4rc1e ptal

PR #1032, opened 2024-09-24
[Builder] Clean up option in config file

emmamarichal:
@simoncozens I tried to make possible the choice of `True` or `False` for the clean up action.
Should solve this: https://github.com/googlefonts/gftools/issues/1014


simoncozens:
Very nice, thank you!

PR #1036, opened 2024-09-30
Check all name records for Mac IDs

### LINKED ISSUE #1035: Increase nameIDs limit in `drop-mac-names` & `drop-superfluous-mac-names` in `fix.py`
Is there a reason for the upper limit of the nameID values to be 255 (in both `drop-mac-names` and `drop-superfluous-mac-names`?

I'm processing several fonts where the the nameIDs in VFs exceed that limit. 

Screenshot 2024-09-27 at 7 11 50 PM

https://github.com/googlefonts/gftools/blob/a2b3549862d9bc7f62b09f84f6c87bc75da14bae/Lib/gftools/fix.py#L688

simoncozens:
In non-VF days we could just look at all name records up to 255. But there's no upper limit these days, so we look at them all. Fixes #1035


m4rc1e:
I think we should just remove mac names completely.

Issue #1045, opened 2024-10-14
Supporting `fontc` as an experimental alternative compiler


cmyr:
I would like to start testing using `fontc` in place of `fontmake` in the gftools builder pipeline, and I'm hoping to discuss how best to do this. Ideally this work would be isolated and non-invasive enough that it could happen in this repo, instead of in a fork.

The main requirement for this work would be some way to pass a flag that would opt us into `fontc` mode. I imagine that this would involve adding some new 'Operations' that can be run in place of the current `BuildOTF` and `BuildVariable` classes, and which would be backed by `fontc`. (An environment variable could be used to pass a path to the `fontc` executable.)

When the user passed the `--experimental-fontc` (or whatever) flag, we would then use the alternative operations instead of the default ones; I'm not sure how best to plumb this through, but I could imagine possibly modifying the `known_operations` or else adding an `operation_overrides` dict or something like that; I imagine this project's maintainers might have a better sense of how best to do this.

Does this make approximate sense? Is there anything else I'm missing?

simoncozens:
There's precedent for using alternative implementations of an operation: see the (now badly-named) `hbsubset` operation which can be switched between `hb-subset` and `pyftsubset`. But that's a bit easier to handle because the command line arguments are deliberately identical; we can just have a `$subsetter` variable in the ninja rule and set it appropriately.

How compatible are fontmake and fontc command lines?

rsheeter:
Thinking "aloud" we are eventually going to need gftools in Rust which makes me wonder if we should begin to spin that up.

cmyr:
> There's precedent for using alternative implementations of an operation: see the (now badly-named) `hbsubset` operation which can be switched between `hb-subset` and `pyftsubset`. But that's a bit easier to handle because the command line arguments are deliberately identical; we can just have a `$subsetter` variable in the ninja rule and set it appropriately.
> 
> How compatible are fontmake and fontc command lines?

They aren't identical but if I know the fontmake args I can try to rewrite them for fontc?

simoncozens:
See notes on above PR.

cmyr:
Opened #1047 to show loosely how I imagine this could work. @simoncozens can you identify any possible problems with that approach?

PR #1046, opened 2024-10-14
Use fontc as alternative gftools.builder compiler

simoncozens:
Implements #1045, if you set `GFTOOLS_COMPILER=fontc` environment variable. Some caveats:

- Any `args` that invoke `--filter` are dropped on the floor. Ideally these should be converted, ie. `--filter FlattenComponentsFilter` should become `--flatten-components`, etc.
- `instantiateUFO` -> `buildTTF` pipelines won't quite work yet as we assume JSON UFO structures in instantiated UFOs, which `fontc` doesn't support. Maybe we could make the JSON-ness dependent on `FontmakeOperationBase.compiler != "fontc"`.

### DIFF
--- a/Lib/gftools/builder/operations/__init__.py
+++ b/Lib/gftools/builder/operations/__init__.py
@@ -5,12 +5,24 @@ import os
 import pkgutil
 import importlib
 import inspect
+import logging
 import sys
 from os.path import dirname
 from tempfile import NamedTemporaryFile
 
 from gftools.builder.file import File
 from gftools.utils import shell_quote
+
+log = logging.getLogger(__name__)
+
+
+# Grossness to allow class properties in Python
+class classproperty:
+    def __init__(self, func):
+        self.fget = func
+
+    def __get__(self, instance, owner):
+        return self.fget(owner)
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


Lib/gftools/builder/operations/__init__.py
```diff
@@ -131,9 +143,24 @@ def validate(self):
 
 
 class FontmakeOperationBase(OperationBase):
+    COMPILER_ENV_KEY = "GFTOOLS_COMPILER"
```
cmyr, line 146: couldn't we just have a different subclass of `OperationBase`, and switch to that dynamically? then we'd avoid junking up this class with fontc stuff, and I could properly handle the arguments...

simoncozens: If you believe it would be cleaner, I would be happy to see it.

cmyr: The only part I'm not sure about is how best to swap the classes in the presence of the CLI flag. Do you have any ideas?

simoncozens: Nope, not really. Once a class is loaded in Python it's unpleasant to try to change its base class. This is why I suggested doing things the way that I did ...

rsheeter: I'm not following why we'd need to change base, we should have two different classes and pick which one to create based on cli. I don't know the gftools codebase, is there something about it that makes that difficult?

simoncozens: If you're not changing base class, then you will need six different classes (one implementation each of `buildTTF`, `buildOTF`, `buildVariable` for fontc and fontbakery). This is *less* elegant than the solution I'm proposing.

rsheeter: I think I would dispute the elegance of writing `if self.compiler ==` :)

PR #1047, opened 2024-10-14
Support for fontc and crater

cmyr:
Here is my attempt to add support in `gftools builder` for running `fontc`, and generally making it easier to use in the context of `fontc_crater`.

- adds the `--experimental-fontc` flag, which, if set:
    - replaces the `BuildVariable`, `BuildTTF`, and `BuildOTF` operations with alternatives that use `fontc`
    - modifies the config so that only _one_ of `buildVariable` and `buildStatic` is `true` (so do not instance variable fonts) and also so that only one of `buildTTF` and `buildOTF` are true (fontc does not currently build CFF outlines, so we turn `buildOTF` off if `buildTTF` is true)
- adds the `--experimental-simple-output=PATH` option, which:
    - if modifies the config so that all built fonts end up in one directory
    - and does the same config pruning as above, so that we can get the same set of outputs with both compilers.

Overall I've tried to make this as noninvasive as possible to the existing code, and I think that has been reasonably successful?

### DIFF
--- a/Lib/gftools/builder/file.py
+++ b/Lib/gftools/builder/file.py
@@ -48,6 +48,12 @@ class File:
         return self.is_glyphs or self.is_ufo or self.is_designspace
 
     @cached_property
+    def is_variable(self) -> bool:
+        return (self.is_glyphs and len(self.gsfont.masters) > 1) or (
+            self.is_designspace and len(self.designspace.sources) > 1
+        )
+
+    @cached_property
     def gsfont(self):
         if self.is_glyphs:
             import glyphsLib
--- a/Lib/gftools/builder/fontc.py
+++ b/Lib/gftools/builder/fontc.py
@@ -11,6 +11,7 @@ from argparse import Namespace
 from pathlib import Path
 from typing import Union
 
+from gftools.builder.file import File
 from gftools.builder.operations.fontc import set_global_fontc_path
 
 
@@ -36,13 +37,22 @@ class FontcArgs:
 
     # update the config dictionary based on our special needs
     def modify_config(self, config: dict):
+        if self.single_source:
+            filtered_sources = [s for s in config["sources"] if self.single_source in s]
+            n_sources = len(filtered_sources)
+            if n_sources != 1:
+                raise ValueError(
+                    f"--exerimental-single-source {self.single_source} must match exactly one of {config['sources']} (matched {n_sources}) "
+                )
+            config["sources"] = filtered_sources
+
         if self.fontc_bin_path or self.simple_output_path:
             # we stash this flag here to pass it down to the recipe provider
             config["use_fontc"] = self.fontc_bin_path
             config["buildWebfont"] = False
             config["buildSmallCap"] = False
             # override config to turn not build instances if we're variable
-            if config.get("buildVariable", True):
+            if self.will_build_variable_font(config):
                 config["buildStatic"] = False
             # if the font doesn't explicitly request CFF, just build TT outlines
             # if the font _only_ wants CFF outlines, we will try to build them
@@ -55,14 +65,14 @@ class FontcArgs:
             config["outputDir"] = str(output_dir)
             config["ttDir"] = str(output_dir)
             config["otDir"] = str(output_dir)
-        if self.single_source:
-            filtered_sources = [s for s in config["sources"] if self.single_source in s]
-            n_sources = len(filtered_sources)
-            if n_sources != 1:
-                raise ValueError(
-                    f"--exerimental-single-source {self.single_source} must match exactly one of {config['sources']} (matched {n_sources}) "
-                )
-            config["sources"] = filtered_sources
+
+    def will_build_variable_font(self, config: dict) -> bool:
+        # if config explicitly says dont build variable, believe it
+        if not config.get("buildVariable", True):
+            return False
+
+        source = File(config["sources"][0])
+        return source.is_variable
 
 
 def abspath(path: Union[Path, None]) -> Union[Path, None]:
--- a/Lib/gftools/builder/fontc.py
+++ b/Lib/gftools/builder/fontc.py
@@ -52,6 +52,10 @@ class FontcArgs:
             config["buildWebfont"] = False
             config["buildSmallCap"] = False
             config["splitItalic"] = False
+            # set --no-production-names, because it's easier to debug
+            extra_args = config.get("extraFontmakeArgs") or ""
+            extra_args += " --no-production-names"
+            config["extraFontmakeArgs"] = extra_args
             # override config to turn not build instances if we're variable
             if self.will_build_variable_font(config):
                 config["buildStatic"] = False
--- a/Lib/gftools/builder/fontc.py
+++ b/Lib/gftools/builder/fontc.py
@@ -14,6 +14,7 @@ from gftools.builder.operations.fontc import set_global_fontc_path
 class FontcArgs:
     simple_output_path: Union[Path, None]
     fontc_bin_path: Union[Path, None]
+    single_source: Union[str, None]
 
     # init with 'None' returns a default obj where everything is None
     def __init__(self, args: Union[Namespace, None]) -> None:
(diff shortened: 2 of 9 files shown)


Lib/gftools/builder/operations/__init__.py
```diff
@@ -149,18 +151,42 @@ def variables(self):
...
+        if use_fontc:
+            self.known_operations["buildVariable"] = FontcBuildVariable
+            self.known_operations["buildTTF"] = FontcBuildTTF
+        print(self.known_operations, file=sys.stderr)
+
+    def get(self, operation_name: str):
+        return self.known_operations.get(operation_name)
+
+
+def get_known_operations() -> Dict[str, OperationBase]:
```
rsheeter, line 175: Operation registry is basically a factory. If we had a different factory for fontc vs dynamicly loaded python, created based on flags, then only the python one would need to do the dynamic module import bit. 

Considering most things use the googlefonts implementation and it always makes the same choices I find myself wondering if there should be a static implementation of that but I suppose that's out of scope.

cmyr: I presume the dynamic loading stuff exists for a reason, and I'm not trying to make any big changes here; just override some of the operations in the presence of a flag.

The rest of the operation types we're happy to share.

cmyr:
Okay I've gone and cleaned this up to the point where I think it is ready to merge, or at least is ready for more considered review. Overall I'm quite happy with how this has gone; I've gotten more or less everything I was hoping for without needing to do anything very invasive.

Lib/gftools/builder/fontc.py
```diff
@@ -0,0 +1,67 @@
+"""functionality for running fontc via gftools
+
+This mostly exists so that we can keep as much of the fontc logic in one place,
+and not need to dirty up anything else.
+"""
```
rsheeter, line 5: Provide a link to the fontc repo, reader might not know what a fontc is.

Lib/gftools/builder/operations/fontc/__init__.py
```diff
@@ -0,0 +1,55 @@
+from pathlib import Path
+from typing import List
+from gftools.builder.operations import OperationBase
+
+FONTC_PATH = "FONTC_PATH_NOT_SET"
```
rsheeter, line 5: _ prefix for a "private" thing.

Suggest default value of None.

Lib/gftools/builder/operations/fontc/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+# should only be called once, from main, before doing anything else. This is a
+# relatively non-invasive way to smuggle this value into FontcOperationBase
+def set_global_fontc_path(path: Path):
+    global FONTC_PATH
+    FONTC_PATH = path
```
rsheeter, line 12: This should fail if the FONTC_PATH is set already and fail if path doesn't point to a file. Maybe even fail if it doesn't point to something executable.

rsheeter (approved):

rsheeter:
Heads up CI is upset with you

Lib/gftools/builder/operations/fontc/__init__.py
cmyr: I chose this default value so that if somehow this ended up not getting set there would be a slightly more informative error message. Is that too clever?

Lib/gftools/builder/operations/fontc/__init__.py
cmyr: I tried doing this logic and then python had some error about setting a global twice? so I think that's actually baked in now?

the check that the file exists happens earlier, at the call site.

rsheeter: Hm, I don't remember there being any rules about multiple writes to globals.

I just tried locally and python happily let me write to a global repeatedly.

cmyr:
@m4rc1e would you like to take a look at this and let me know if it seems sane / if there's a simpler way I might achieve any of my goals, here?

Lib/gftools/builder/operations/__init__.py
```diff
@@ -150,17 +151,46 @@ def variables(self):
...
+    def get(self, operation_name: str):
+        if self.use_fontc:
+            if operation_name == "buildVariable":
+                # if we import this at the top level it's a circular import error
+                from .fontc.fontcBuildVariable import FontcBuildVariable
+
+                return FontcBuildVariable
+            if operation_name == "buildTTF":
+                from .fontc.fontcBuildTTF import FontcBuildTTF
+
```
anthrotype, line 168: I think you forgot a `return FontcBuildTTF`...

Lib/gftools/builder/operations/fontc/fontcBuildOTF.py
```diff
@@ -0,0 +1,8 @@
+from gftools.builder.operations.fontc import FontcOperationBase
+
+
+class FontcBuildTTF(FontcOperationBase):
```
anthrotype, line 4:
```suggestion
class FontcBuildOTF(FontcOperationBase):
```

Lib/gftools/builder/file.py
```diff
@@ -47,6 +47,12 @@ def is_designspace(self):
     def is_font_source(self):
         return self.is_glyphs or self.is_ufo or self.is_designspace
 
+    @cached_property
+    def is_variable(self) -> bool:
+        return (self.is_glyphs and len(self.gsfont.masters) > 1) or (
```
m4rc1e, line 52: This may be too simple since I have a seen a family that has 4 masters and 4 instances which are not MM compatible so it could only export 4 static fonts.

Something like `self.is_glyphs and len(self.gsfont.masters) < len(self.gsfont.instances)` but it still isn't 100% reliable.

Lib/gftools/builder/operations/fontc/__init__.py
```diff
@@ -0,0 +1,61 @@
...
+        if args:
+            vars["args"] = rewrite_fontmake_args_for_fontc(args)
+
+        return vars
+
+
+def rewrite_fontmake_args_for_fontc(args: str) -> str:
+    out_args = []
+    arg_list = args.split()
+    # reverse so we can pop in order
```
m4rc1e, line 31: Nit. I'm probably being dense here but couldn't skip the reverse part and just keep popping the first element by using `args.pop(0)`,

Lib/gftools/builder/file.py
anthrotype: perhaps `maybe_variable` is a better name for the method. But I think the gftools config provides an even stronger hint as to whether a font is supposed to be variable: if the font dev asks to build a variable font, then it must be variable; if it disables building the VF (is it `buildVariable: false`?) then one should treat it as not variable.

m4rc1e (commented):
Nice work @cmyr. I'm not the original author of builder 2 so I'll let @simoncozens have a look before merging.

I've tested it on a few families and it's working well for its intended purposes.

I have no problems with this PR being merged. However, once fontc and its sibling libs/tools have matured, it would be great to drop the experimental flags and allow the user to specify which compiler they would like in the config file. This will drop the need to have a seperate `fontc_args` argument for the `GFBuilder` class, which should make the code a touch tidier.

My only slight hesitation is:

> modifies the config so that only one of buildVariable and buildStatic is true 

Currently, a config's default settings will export otf, ttf and var and make web versions for all formats. It's sad to see this go but since it is an experimental feature used solely by you folks, I'm ok with this.

m4rc1e (approved):

Lib/gftools/builder/file.py
simoncozens: > But I think the gftools config provides an even stronger hint as to whether a font is supposed to be variable: if the font dev asks to build a variable font, then it must be variable; if they disables building the VF (is it buildVariable: false?) then one should treat it as not variable.

I'm not sure I understand this. `buildStatic`, `buildVariable`, `buildOTF`, `buildWebfont` etc. are parameters about *targets*, not about *sources*. "I don't want you to build a variable font today" does not mean "this is not a variable font".

PR #1062, opened 2024-11-12
[fontc] Only crater needs unique build file names

cmyr:
When running in the context of crater, gftools builder may be invoked multiple times in parallel, which can cause the generated ninja.build files to overwrite each other.

To fix this, I implemented a scheme where we generate a unique name for the generated build file, so that multiple could exist and be in use for the same project, at the same time.

These files can pile up, though; I had initially thought they were cleaned up by default, but that doesn't seem to be the case.

This patch makes it so we only use this new naming scheme if we're running for crater, and uses the old names otherwise.

It also makes it so we always set the cleanup flag if we're running for crater.

### DIFF
--- a/Lib/gftools/builder/fontc.py
+++ b/Lib/gftools/builder/fontc.py
@@ -10,6 +10,7 @@ dirty up everything else.
 from argparse import Namespace
 from pathlib import Path
 from typing import Union
+import time
 
 from gftools.builder.file import File
 from gftools.builder.operations.fontc import set_global_fontc_path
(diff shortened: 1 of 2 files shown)


Lib/gftools/builder/fontc.py
```diff
@@ -35,6 +36,15 @@ def __init__(self, args: Union[Namespace, None]) -> None:
     def use_fontc(self) -> bool:
         return self.fontc_bin_path is not None
 
+    def build_file_name(self) -> str:
+        if self.fontc_bin_path or self.simple_output_path:
+            # if we're running for fontc we want uniquely named build files,
+            # to ensure they don't collide
+            return f"build-{time.time_ns()}.ninja"
```
m4rc1e, line 43: Out of curiosity, why nanoseconds over something like a uuid?

PR #1065, opened 2024-11-14
[add-font] Continue if license detection fails

simoncozens:
This allows people to use `gftools-add-font` to create METADATA.pb files outside of the fonts.google.com infrastructure.

### DIFF
--- a/Lib/gftools/scripts/add_font.py
+++ b/Lib/gftools/scripts/add_font.py
@@ -152,7 +152,6 @@ def _MakeMetadata(args, is_new):
 
     first_file = file_family_style_weights[0].file
     old_metadata_file = os.path.join(args.directory, "METADATA.pb")
-    font_license = fonts.LicenseFromPath(args.directory)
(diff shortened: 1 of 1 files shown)


Lib/gftools/scripts/add_font.py
```diff
@@ -214,7 +213,12 @@ def _MakeMetadata(args, is_new):
             ):
                 metadata.primary_script = script
 
-    metadata.license = font_license
+    try:
+        font_license = fonts.LicenseFromPath(args.directory)
+        metadata.license = font_license
+    except ValueError:
+        print("WARNING: Could not determine license for %s" % args.directory)
+        print("Continuing without setting license.")
```
m4rc1e, line 221: Thoughts on setting it to OFL if this fails?

simoncozens: For the use case I have in mind, we very much do not want this to default to OFL.

PR #1067, opened 2024-11-20
Disable JSON structure when using fontc

simoncozens:
Normally, gftools-builder uses a small optimisation - saving intermediate UFOs in ufoLib2's "JSON structure" which is faster to serialize and parse. However, we want to turn off this optimisation when running with `fontc` as it does not support this (undocumented, non-standard) UFO representation.

### DIFF
--- a/Lib/gftools/builder/__init__.py
+++ b/Lib/gftools/builder/__init__.py
@@ -37,6 +37,7 @@ def edge_with_operation(node, operation):
 class GFBuilder:
     config: dict
     recipe: Recipe
+    fontc_args: FontcArgs
 
     def __init__(
         self,
@@ -60,6 +61,7 @@ class GFBuilder:
         else:
             self._orig_config = yaml.dump(config)
             self.config = config
+        self.fontc_args = fontc_args
         fontc_args.modify_config(self.config)
 
         self.known_operations = OperationRegistry(use_fontc=fontc_args.use_fontc)
@@ -150,6 +152,9 @@ class GFBuilder:
         glyph_data = self.config.get("glyphData")
         if glyph_data is not None:
             glyph_data = glyph_data
+        ufo_structure = "json"
+        if self.fontc_args.use_fontc:
+            ufo_structure = "package"
         FontProject().run_from_glyphs(
             str(source.resolve()),
             **{
@@ -157,7 +162,7 @@ class GFBuilder:
                 "output_dir": directory,
                 "master_dir": directory,
                 "designspace_path": output,
-                "ufo_structure": "json",
+                "ufo_structure": ufo_structure,
                 "glyph_data": glyph_data,
             },
         )
--- a/Lib/gftools/builder/operations/__init__.py
+++ b/Lib/gftools/builder/operations/__init__.py
@@ -173,6 +173,11 @@ class OperationRegistry:
 
                 return FontcBuildOTF
 
+            if operation_name == "addSubset":
+                from .fontc.fontcAddSubset import FontcAddSubset
+
+                return FontcAddSubset
+
         return self.known_operations.get(operation_name)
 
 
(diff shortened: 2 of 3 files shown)


simoncozens:
@cmyr With this I can build Noto fonts using fontc, except of course all the static generation is gone.

simoncozens:
With --experimental-fontc, building Noto Sans Arabic now takes 1m 52s on my machine, compared to 1m 24 without, presumably because of the slower UFO marshalling.

I would be very interested in changing the `fontc` directory to a `rust` directory and using it to host Rust implementations of other operations, such as using `triangulate` to instantiate the UFOs, `fontspector --hotfix` instead of `gftools-fix` etc.

Lib/gftools/builder/fontc.py
```diff
@@ -69,13 +69,10 @@ def modify_config(self, config: dict):
             extra_args = config.get("extraFontmakeArgs") or ""
             extra_args += " --no-production-names --drop-implied-oncurves"
             config["extraFontmakeArgs"] = extra_args
-            # override config to turn not build instances if we're variable
-            if self.will_build_variable_font(config):
-                config["buildStatic"] = False
```
cmyr, line 74: Are you accomplishing this another way, or are you building statics? The current `fontc` mode is very much limited in focus to supporting the crater project, which lets us run both compilers and compare outputs, and I would like to avoid doing extra work in that case because it already takes quite a long time to run.

Maybe we could keep this, but have it be gated on the `--fontc-experimental-single-source` argument?

Lib/gftools/builder/recipeproviders/googlefonts.py
```diff
@@ -333,8 +333,7 @@ def build_a_static(self, source: File, instance: InstanceDescriptor, output):
         steps = [
             {"source": source.path},
         ]
-        # if we're running fontc we skip conversion to UFO
-        if not source.is_ufo and not self.config.get("use_fontc", False):
```
cmyr, line 337: if we remove this I believe we can also remove the place where we add the `use_fontc` flag to the config? it felt very hacky, and I believe it was only used in this one place.

Lib/gftools/builder/__init__.py
```diff
@@ -37,6 +37,7 @@ def edge_with_operation(node, operation):
 class GFBuilder:
     config: dict
     recipe: Recipe
+    fontc_args: FontcArgs
```
cmyr, line 40: do you actually use this at all later? One of the ideas of the original design was that this could be a transient thing that was just used to set some config flags and then could be thrown away, and the rationale for that was that I wanted to avoid polluting the project with stuff that I would anticipate wanting to rip out later.

If there's a reason to keep it sure, but otherwise it's probably not worth?

cmyr (approved):
I'm glad you're finding this helpful! My only real desire with the fontc-supporting code is that it continues to support the needs of crater as well as it does now, which means really doing the minimal amount of work possible to get  us the artifacts we're comparing, which are one generated binary per run. So long as we can continue doing that as needed, then this looks good.

Do ping me before publishing please just so I can do a quick crater run locally and make sure everything is working as expected, none of this fontc stuff is really tested in CI...

Lib/gftools/builder/__init__.py
simoncozens: Yes, see `glyphs_to_ufo` below - if we're using fontc, then UFOs should be stored in package structure, not JSON structure.

Lib/gftools/builder/recipeproviders/noto.py
```diff
@@ -277,7 +277,7 @@ def build_a_static(self, source, instance, output):
                     "instance_name": instance.name,
                     "target": "full-designspace/instance_ufos/"
                     + os.path.basename(instance.filename)
-                    + ".json",
+                    + ("" if self.builder.fontc_args.use_fontc else ".json"),
```
simoncozens, line 280: @cmyr Another place where we need to know if we're running fontc or not.

PR #1076, opened 2025-01-17
[fontc] Handle more fontmake args in fontc mode

cmyr:
Fleshing these out as they arise; noticed a few more causing failures on crater.


anthrotype:
it would be nice to get a release containing this patch, so we can use on fontc_crater, thanks! 🙏

m4rc1e:
Coming up

m4rc1e:
Done, https://github.com/googlefonts/gftools/releases/tag/v0.9.77

anthrotype:
Thanks!

Issue #469, opened 2022-01-19
[builder] move the mac names specific to VF to platform 3?


RosaWagner:
I transmit here the conversion we had in the group chat to keep an archive and have more opinions:

Original question:
Is there a reason why we have nameID 25 + custom names linked to STAT and FVAR (25* to 27*) in platform 1 instead of platform 3 with the rest of the name tables?

Adam said:
> As we know, variable "OpenType" fonts have been working in Mac OS since Mac OS 7.5 released in 1994, and if packaged as TTF, since the first version of Mac OS X published in 2001. This is because TueType-flavored variable fonts are backwards-compatible with TrueType GX Variations. But until Mac OS X 10.13 High Sierra (2017) or 10.14 Mojave (2018), not sure now, the predefined instances in the fvar table only worked if their names were registered in the Mac classic PID 1 EID 0 name table branch.

> Even though macOS started preferring the 3.1 name entires for other names, Apple’s TT GX code was dormant for years so nobody updated that part, until, I think Mojave. There is still some portion of users of Mac OS X Sierra and High Sierra, especially since older Apple hardware is actually surprisingly resilient, and still yields good prices on secondary markets. In some countries, a seriously older Mac with old Mac OS X is still a machine that is much in demand.

So main concern is backward compatibility. 

Although, Dave recalled that these OS don't support VF so well.

Also Olli (@moontypespace) told me by email:
> Even apple’s SFCompact.ttf system variable font does not contain Mac name table entries anymore.

cc @vv-monsalve @davelab6 @twardoch @m4rc1e

moontypespace:
@RosaWagner Thanks for bringing this up. I would like to add to this conversation, that Dave Opstad (http://luc.devroye.org/fonts-43562.html) told me, we should not produce fonts with Mac entries anymore. Dave helped develop the TrueType font format at Apple (for many years) and is one of the authors of Apple's GX.

felipesanches:
@moontypespace has provided a new FontBakery check related to this discussion at https://github.com/googlefonts/fontbakery/pull/3545

If we ever consider this a generic check, then we should consider incorporating it on FontBakery's 'universal' profile, but I'd like to get some feedback from you all before we do so. So, for now. It will stay on the newly created **Fontwerk** profile.

Feel free to open an issue on the FontBakery issue tracker if any of you feel that this should be in `universal`.

[...]

PR #1077, opened 2025-01-22
Do not add Mac names when populating the STAT table

Hoolean:
See #469

As of the latest Font Bakery release, [the `no_mac_entries` check has been promoted to the Universal profile](https://github.com/fonttools/fontbakery/blob/main/CHANGELOG.md#0130a0-2024-sep-13); by default, gftools and axisregistry insert Mac names when populating `STAT` and `fvar`, and so boilerplate gftools repositories are FAIL'ing this check out-of-the-box.

This PR makes a partial fix, by preventing Mac names from being added to `STAT`.

This is the right approach for the particular project I am looking at, but I would like to confirm that this is the right fix - as opposed to, e.g., making the Mac names opt-out - to make sure that this will not have unintended consequences for static TTFs destined for legacy environments, or similar.

### DIFF
--- a/Lib/gftools/stat.py
+++ b/Lib/gftools/stat.py
@@ -127,4 +127,4 @@ def gen_stat_tables_from_config(stat, varfonts, has_italic=None, locations=None)
             if filename not in locations:
                 raise ValueError("Filename %s not found in locations" % filename)
             locations = locations[filename]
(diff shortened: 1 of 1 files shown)


Hoolean:
cc @simoncozens

m4rc1e:
Thanks!

PR #1079, opened 2025-01-28
add -q, --quiet and --no-backup options to fix_nonhinting.py

### LINKED ISSUE #1078: fix-nonhinting leaves debris
`gftools fix-nonhinting`  is a simple command with two arguments: an input font and an output font:

```
$❯ gftools fix-nonhinting -h
usage: gftools [-h] fontfile_in fontfile_out
```

#### Why is it necessary to make a backup of the input font? 
Although the tool makes no changes to the input font file, it creates a backup of it in the same directory.

**Possible options:**
- remove that part of the logic entirely
- provide a `--no-backup` option to disable the logic

#### The command `print`s a lot of relatively useless output:
- "backup font saved" message
- dump of input font's gasp table if present, otherwise a message that the gasp table wasn't present
- dump of input font's prep table if present, otherwise a message that the prep table wasn't present
- dump of output font's gasp table
- dump of output font's prep table
- a "saved" message.

At a minimum that is 12 unnecessary lines of output text put into the build log and might, if the input font was already hinted, be a lot more.

**Possible options:**
- quiet those by default and enable with a `-v` option for those who want them
- leave them enabled by default and quiet them with a `-q` option

bobh0303:
doesn't change default behavior but fixes #1078

### DIFF
--- a/Lib/gftools/scripts/fix_nonhinting.py
+++ b/Lib/gftools/scripts/fix_nonhinting.py
@@ -56,48 +56,57 @@ from gftools.fix import fix_unhinted_font
 parser = ArgumentParser(description=__doc__, formatter_class=RawTextHelpFormatter)
 parser.add_argument("fontfile_in", nargs=1, help="Font in OpenType (TTF/OTF) format")
 parser.add_argument("fontfile_out", nargs=1, help="Filename for the output")
+parser.add_argument(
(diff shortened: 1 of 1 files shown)


m4rc1e:
Thanks you!

Mind running black on the file to fix the linting issues?

bobh0303:
> Mind running black on the file to fix the linting issues?

Done  (with apologies - I should have known to do that)

PR #1081, opened 2025-02-04
[fontc] Be more defensive during cleanup

cmyr:
Specifically, we don't want to delete the instance_ufos directory if it looks like another instance of gftools might be running in the same location, since it might contain files that are being used.

(The disapearance of this directory has been causing intermittent crater failures.)


anthrotype (approved):

m4rc1e:
Thank Colin for the pr and Cosimo for the review. Would you both like a new release?

cmyr:
@m4rc1e no rush but eventually that would be nice :)

PR #1084, opened 2025-02-19
[gftools-build-vtt] change script name

simoncozens:
gftools installs a script called `gftools-build-vtt`, but it currently has the wrong name for the Python module providing that script, so it doesn't work. I'm sort of surprised nobody else noticed this.


felipesanches:
Thanks, @simoncozens

This should fix 12 targets in **fontc_crater** once we get a new gftools release.
(https://github.com/googlefonts/fontc_crater/issues/12)

guidoferreyra:
Is it planned to have a new release of gf-tools soon?

PR #1089, opened 2025-03-08
Raise an error when an unknown fontmake filter is provided to fontc

Hoolean:
My understanding is that this was the original intention of the code, but that the fall-through case performs the task of skipping too. (cc @cmyr would benefit from your know-how here, is this indeed what we want to support? :mag_right: )

Merging would lead to additional fails on crater, but in at least some cases (e.g. the Jersey family) for families that have a low % match already owing to an unsupported filter; this would become more explicit.

<sub>Note: this is a personal contribution independent of my employer, and so I've submitted from a fork under my personal profile and email to make this distinction</sub>

### DIFF
--- a/Lib/gftools/builder/operations/fontc/__init__.py
+++ b/Lib/gftools/builder/operations/fontc/__init__.py
@@ -43,7 +43,7 @@ def rewrite_one_arg(args: List[str]) -> str:
         # this means 'retain filters defined in UFO', which... do we even support
         # that in fontc?
         if filter_ == "...":
-            pass
+            return ""
         elif filter_ == "FlattenComponentsFilter":
             return "--flatten-components"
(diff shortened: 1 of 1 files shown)


cmyr (approved):
Was I misunderstanding the semantics of `pass`? The intention was that we would raise an error when we saw an unknown filter, so that we could explicitly decide how we wanted to hande them.

simoncozens:
> Was I misunderstanding the semantics of `pass`? The intention was that we would raise an error when we saw an unknown filter, so that we could explicitly decide how we wanted to hande them.

In your original code, `pass` would do nothing when we see an unknown filter, but because we had matched the `if next_ == "--filter":` branch in the outer if-elif tree, the outer `else` path wouldn't be taken, we'd drop to the end and `return ""`.

PR #1092, opened 2025-04-03
builder: add ability to include avar2 mappings

m4rc1e:
This pr allows users to specify avar2 mappings in their builder files e.g:
```
avar2:
  MavenPro[wght].ttf:
  - in:
      wght: 400.0
    out:
      wght: 500.0
  - in:
      wght: 500.0
    out:
      wght: 700.0
```


m4rc1e:
@simoncozens I'm not very familiar with your builder so I'd be very grateful if you could review this pr.

simoncozens:
I think we're missing a `Lib/gftools/builder/operations/buildAvar2.py`

m4rc1e:
Oh my, I can get copilot to review this... why not!

Lib/gftools/builder/recipeproviders/googlefonts.py
```diff
@@ -268,6 +282,18 @@ def build_STAT(self):
...
+    def build_Avar2(self):
+        all_variables = [x for x in self.recipe.keys() if x.endswith("ttf")]
+        if len(all_variables) > 0:
+            last_target = all_variables[-1]
+            args = {"args": self.avar2file.name}
+            build_avar2_step = {
+                "postprocess": "buildAvar2",
+                **args,
+            }
+            self.recipe[last_target].append(build_avar2_step)
```
simoncozens, line 294: This will only work when there is one VF produced by the builder (i.e. a Regular, but not Regular/Italic VFs). This is because the STAT step applies across all VFs produced, whereas this one works on a per-VF basis. You probably want to call it in the `build_a_variable` step instead of in the `build_all_variables` step, and do it to an individual VF rather the last one.

Lib/gftools/builder/recipeproviders/googlefonts.py
```diff
@@ -268,6 +282,18 @@ def build_STAT(self):
                 build_stat_step["needs"] = other_variables
             self.recipe[last_target].append(build_stat_step)
 
+    def build_Avar2(self):
```
simoncozens, line 285: While I'm being pernickety, let's have standard snake case method names (`build_avar2`). (`build_STAT` is just about OK because that's the table name.)

Lib/gftools/builder/recipeproviders/googlefonts.py
```diff
@@ -268,6 +282,18 @@ def build_STAT(self):
...
+        all_variables = [x for x in self.recipe.keys() if x.endswith("ttf")]
+        if len(all_variables) > 0:
+            last_target = all_variables[-1]
+            args = {"args": self.avar2file.name}
+            build_avar2_step = {
+                "postprocess": "buildAvar2",
+                **args,
```
simoncozens, line 292: Overcomplicated, just put args inline in the dict.

simoncozens (changes requested):

Lib/gftools/builder/recipeproviders/googlefonts.py
m4rc1e: Thanks. I've struggled to implement your feedback in `build_a_variable` since the func's target is a temp file. `gftools gen-avar2` uses the font's filename in order to fetch the relevant mapping. Since `gftools gen-avar2` allows multiple ttfs, I'm still able to call it from `build_all_variables` and it will process all the ttfs. Hopefully my changes should be better than this explanation,

m4rc1e: I've already tested it on a family with both Roman and Italic sources and it works.

m4rc1e:
Simon was happy so I'm going to merge.

PR #1093, opened 2025-04-03
Use Rust for fontspector/diffenator

simoncozens:


m4rc1e (approved):
LGTM.

It would be nice if we could also install the rust tools by doing something like `pip install gftools[qa-rust]`

PR #1106, opened 2025-05-13
[builder] add --experimental-extra-arg flag to pass down extra arguments to compilers

anthrotype:
In ttx_diff.py we'd like control whether to pass the `--no-production-names` flag down to fontc and fontmake or not..

Which means, renaming glyphs to their final production names by default, but sometimes when useful keep original human-readable glyph names for debugging.

I added a catch-all flag which means we won't need extra flags if we'd like to support any additional options.

### DIFF
--- a/Lib/gftools/builder/fontc.py
+++ b/Lib/gftools/builder/fontc.py
@@ -27,6 +27,7 @@ class FontcArgs:
         self.simple_output_path = abspath(args.experimental_simple_output)
         self.fontc_bin_path = abspath(args.experimental_fontc)
         self.single_source = args.experimental_single_source
+        self.extra_args = args.experimental_extra_args or []
         if self.fontc_bin_path:
             if not self.fontc_bin_path.is_file():
                 raise ValueError(f"fontc does not exist at {self.fontc_bin_path}")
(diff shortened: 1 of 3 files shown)


cmyr (approved):
Thanks cosimo, not pretty but it should do the trick.

Lib/gftools/builder/fontc.py
```diff
@@ -65,9 +66,8 @@ def modify_config(self, config: dict):
             config["cleanUp"] = True
             # disable running ttfautohint, because we had a segfault
             config["autohintTTF"] = False
-            # set --no-production-names, because it's easier to debug
             extra_args = config.get("extraFontmakeArgs") or ""
```
simoncozens, line 69: Isn't this already a way to get extra args to the compiler?

anthrotype: well, that's the way if you start with a config. here ttx_diff wants to override the user config and to pass down some extra command line options

anthrotype: this is where gftools builder gets called from ttx_diff.py

https://github.com/googlefonts/fontc/blob/8ffba4caa50c68b4fc9916d5e77461e98b9e6f83/resources/scripts/ttx_diff.py#L275-L299

we want to inject some extra flag for fontc/fontmake. Are you suggesting that we modify the input yaml before passing it to gftools?

simoncozens: Well, that's how I would do it - it would avoid special-case stuff like this. But either way works. So long as you had considered it and thought this was easier...

PR #1111, opened 2025-05-19
Fix article media

m4rc1e:


Lib/gftools/scripts/fix_article.py
```diff
@@ -0,0 +1,195 @@
...
+    # If an image is already a jpg, we don't want to compress it further
+    # since it'll create compression artifacts.
+    if img_size > MAXSIZE_RASTER and fp.suffix in [".jpg", ".jpeg"]:
+        raise ValueError(
+            f"Image '{fp}' is too large for a jpg: {img_size} > {MAXSIZE_RASTER}."
+        )
+    print("Converting '{fp}' to jpg")
+    img = img.convert("RGB")
+    new_fp = fp.with_suffix(".jpg")
+    img.save(new_fp, "JPEG", quality=85)
```
simoncozens, line 55: Maybe do another file size check here - we may not have fixed the problem.

PR #1113, opened 2025-05-20
[fontc] turn fontmake's --filter DecomposeComponents into fontc's --decompose-components

anthrotype:
to be used in conjunction with https://github.com/googlefonts/fontc/pull/1458


cmyr (approved):
👍

PR #1124, opened 2025-06-22
[servers] Don’t die on 404

simoncozens:
Sometimes a server claims to have a family, but it doesn't, so we can't download it. If we fail to download it, don't crash, but just return emptiness.


simoncozens:
(To reproduce the problem, delete your `.gf_server_data.json` and try recreating it.)

PR #1150, opened 2025-10-17
avar2 to avar1 added

m4rc1e:


Lib/gftools/scripts/avar2_to_avar1.py
```diff
@@ -0,0 +1,115 @@
...
+            source.name = "_".join(f"{axis_order[i]}-{v}" for i, v in enumerate(combo))
+            source.familyName = ttfont["name"].getBestFamilyName()
+            source.filename = f"{tmpdir}/{source.name}.ttf"
+            source.location = {axis_names[i]: v for i, v in enumerate(combo)}
+            ds.sources.append(source)
+            coords = {axis_order[i]: v for i, v in enumerate(combo)}
+            partial = instancer.instantiateVariableFont(ttfont, coords)
+            partial.save(source.filename)
+
+        inst = InstanceDescriptor()
```
m4rc1e: We should just take all the instances from the avar2 font...duh!

PR #1155, opened 2025-10-30
Remove an unnecessary include directory argument

shimon1024:
When UFO1 and UFO2 are put in different directories and UFO2's directory is specified as an include directory, the inclusion by UFO1 fails because UFO1 and UFO2 share the specified include directory.

This PR fixes [the CI for Noto Sans Mongolian](https://github.com/notofonts/mongolian/actions/runs/18217347321/job/51869384583).


shimon1024:
Thank you!

Issue #1154, opened 2025-10-24
pygit2 should not be pinned to old 1.16.0 version in pyproject.toml


anthrotype:
it makes it harder to install gftools on Python 3.14, as it wants to build from source and requires libgit2 and a compiler.

pygit2 DOES have wheels for 3.14 already, see https://pypi.org/project/pygit2/#files

PR #1160, opened 2025-11-18
drop 3.9, require 3.10+; test on 3.14

anthrotype:
python 3.9 is EOL

Fix https://github.com/googlefonts/gftools/issues/1154

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -58,9 +58,17 @@ jobs:
         run: choco install ffmpeg -y
         shell: pwsh
 
-      - name: Install packages
+      # Temporarily skip QA extra on Python 3.14 until skia-python provides wheels for 3.14.
+      # The qa extra pulls in diffenator2 which depends on blackrenderer[skia] which in turn
+      # depends on skia-python. Building skia-python from source fails on CI.
+      # See: https://github.com/googlefonts/gftools/pull/1160
+      - name: Install packages (with QA)
+        if: matrix.python-version != '3.14'
         run: |
-          pip install '.[qa]'
+          pip install '.[qa,test]'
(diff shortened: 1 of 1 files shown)


khaledhosny:
Any progress with this? I’m getting CI build breakage because conflicting pygit2 requirements, which I think is related to this.

(I really wish if the builder was split into its own repository, since many of the requirements pulled by gftools are not related to the builder, but it is often the only part people need to use.)

anthrotype:
i think the remaining issue for supporting 3.14 is that diffenator2 depends on `blackrenderer[skia]` (for rendering COLRv1 fonts I believe) which pulls in skia-python and the latter doesn't have prebuilt wheels for 3.14 yet and builindg from source fails on the gftools CI (see above).

I suggested @m4rc1e to try and see if using pycairo (which does provide wheels for 3.14) as a backend for `blackrenderer[cairo]` in diffenator2 could fix this.

Currently if you want to use gftools builder you have to stick to python 3.13.

anthrotype:
sorry, correction. gftools only depends on diffenator2 (and thus on skia-python) when you pip install the `gftools[qa]` extra. That's what the CI is doing and why it is failing. But if one wants to install gftools to only use the core builder on Python 3.14, just un-pinning the pygit2 dependency as I did here is sufficient. So I think we should merge this and somehow mute the 3.14 build failure related to diffenator2/skia-python

anthrotype:
The test now pass. I think it's better to merge this to unblock installing the core of gftools (builder, without the "qa" extra) for Python 3.14.

m4rc1e:
Thanks Cosimo!

Will merge this and see if I can port diff2 from Skia to Cairo

PR #1165, opened 2026-01-23
chore(deps): "setuptools" is not required at runtime

a-detiste:
Hi,

We applied this patch on Debian today:

https://salsa.debian.org/fonts-team/gftools/-/commit/4c607d70fda6ddcc65a77e2c472b8313d5a2d77c


simoncozens:
Thanks for this, @a-detiste! (No particular thanks to the other guy, though.)

PR #1166, opened 2026-01-29
Builder: Add fontsetter

m4rc1e:
I need to adjust a few values in Crispy and realised the builder isn't hooked up to fontsetter.


m4rc1e:
Going to merge this since I need to get Crispy out the door. We can always revert and improve if needed.

simoncozens:
A belated LGTM anyway. Sorry!

PR #1169, opened 2026-02-12
SubsetMerger: check extraneous donor axes are at their default location

RickyDaMa:
For example, I was merging a font with a GRAD axis into a font that doesn't have a GRAD axis. Because of the Designspace's ordering, I was getting a non-zero GRAD merged in, making the appearance of the two sources mismatch. By ensuring extra axes like GRAD are at their default locations, we mitigate unexpected sub-optimal matching.

Per the comment about this method's intentional permissiveness, this has no impact on single-master donors getting merged in.

I'll get a unit test thrown together shortly.

### DIFF
--- a/Lib/gftools/subsetmerger.py
+++ b/Lib/gftools/subsetmerger.py
@@ -255,9 +255,10 @@ def is_compatible(
 ) -> bool:
     input_userspace_location = input_descriptor.userspace_location
     my_userspace_location = descriptor.userspace_location
-    common_axis_tags = set(input_userspace_location.keys()).intersection(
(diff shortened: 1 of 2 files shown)


simoncozens:
This is great, but we have fontmerge now and I would prefer to invest in that...

RickyDaMa:
For sure, but I'm migrating projects to current stable Google Fonts template / `gftools`. Unless the `addSubset` operation is going to move to Rust/fontmerge in the very near future, then I'd appreciate it if this change could be accepted and released 🙏

PR #1173, opened 2026-02-20
builder: restore previous googlefonts recipe

### LINKED ISSUE #1172: gftools builder crashes with TypeError in ninja_syntax.py when processing nested list variables (e.g., splitItalic: true)
While assisting @RosaWagner with investigating https://github.com/zalando/sans/actions/runs/21861926252/job/63093057085?pr=42, I found an error in gftools' treatment of nested list variables. 

## Description
When running gftools builder on a config.yaml file that includes operations generating complex metadata fixes—such as splitItalic: true or - operation: fix—the build process crashes while attempting to generate the Ninja build file.

The builder is internally passing a Python list of lists (representing the table fixes) directly to the Ninja writer. Because ninja-syntax expects a flat list of strings to join, it throws a TypeError when it encounters the nested lists.

## Proposed Resolution
The gftools builder needs to intercept these complex internal data structures and safely serialize them into JSON strings before passing them to the Ninja writer payload.

-- 
I was able to resolve this locally via modifying the `ninja_syntax.py` file to serialize the data, but the fix should happen in gftools directly.

```
        if isinstance(value, list):
            try:
                value = ' '.join(filter(None, value))
            except TypeError:
                # Bypass gftools bug: safely serialize nested lists into a quoted JSON string
                import json
                import shlex
                value = shlex.quote(json.dumps(value))
```

---

Gemini suggests modifying: `https://github.com/googlefonts/gftools/blob/main/Lib/gftools/builder/operations/__init__.py`

```
@property
    def variables(self):
        return {k: v for k, v in self.original.items() if k != "needs"}
```

to 

```
@property
    def variables(self):
        import json
        import shlex
        
        vars_dict = {}
        for k, v in self.original.items():
            if k == "needs":
                continue
            # If the value is a complex structure (like a list of lists), 
            # safely serialize it so ninja-syntax doesn't crash on .join()
            if isinstance(v, (list, dict)) and any(isinstance(i, (list, dict)) for i in (v if isinstance(v, list) else v.values())):
                vars_dict[k] = shlex.quote(json.dumps(v))
            else:
                vars_dict[k] = v
        return vars_dict
```

m4rc1e:
My work on adding the fontsetter, #1166 broke the Italic VF split. This PR restores the googlefonts recipe provider to the previous implementation in a4d25179f5d4d794f07e86653821bb8dd1294ed6.

Fixes #1172

### DIFF
--- a/Lib/gftools/builder/recipeproviders/googlefonts.py
+++ b/Lib/gftools/builder/recipeproviders/googlefonts.py
@@ -122,14 +122,6 @@ class GFBuilder(RecipeProviderBase):
         else:
             self.fvarInstancesFile = None
 
-        if "fontsetter" in self.config:
-            self.fontsetterfiles = {}
-            for font in list(self.config["fontsetter"].keys()):
-                tmp_file = NamedTemporaryFile(delete=False, mode="w+")
-                self.config["fontsetter"][font] = self.config["fontsetter"][font]
-                yaml.dump(self.config["fontsetter"][font], tmp_file)
-                self.fontsetterfiles[font] = tmp_file
-
         # Find variable fonts
         self.recipe = {}
         self.build_all_variables()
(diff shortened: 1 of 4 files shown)


m4rc1e:
Let me add a test so this doesn't happen again

PR #1175, opened 2026-02-26
Add gftools-what-version

m4rc1e:
Many families don't pin their dependencies so it makes reproducing a working build difficult. This PR will check which version of gftools can build the family

Still mega WIP.

### DIFF
--- a/Lib/gftools/scripts/what_version.py
+++ b/Lib/gftools/scripts/what_version.py
@@ -1,0 +1,220 @@
+"""Find which versions of gftools can successfully build a font project."""
+
+import argparse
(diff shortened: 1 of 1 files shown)


madig:
(Can you correlate date of latest repo commit and gftools release?)

m4rc1e:
Interesting. I find designers don't update gftools often so also checking the first commit may make a lot of sense.

madig:
Not necessarily the first commit, but consider how whatever the current state of the repo is, it's unlikely to have used a gftools released after the commit you're looking at.

PR #1176, opened 2026-02-26
`addSubset`: support glyph name list/files

RickyDaMa:
Adds two YAML keys under the `subsets` object: `glyphNames` (a list of glyph names) and `glyphNamesFile` (a file containining newline delimited glyph names, comments supported/trimmed). I wish I'd realised in my previous PRs that snake_case was not the naming convention being used in the rest of the YAML document.

Pulls the glyph name reading code I wrote for `exclude_glyphs_file` into a utils function so I can use it for the included glyphs too.

Puts a lil' type hint on `subset(s)` through the codebase.

I also took the liberty of introducing a couple classes to improve the readability of `prepare_minimal_subsets` to try and keep the code more readable, instead of adding a new field into the already nasty-looking `incl_excl_by_donor`. These could be removed though if they're undesirable.

Side note: there's no practical reason why these two conditionals need to be exclusive, and it means if you include both `name` and `ranges` on one subset entry the ranges get silently dropped. Perhaps I could change the `elif` to an `if`?

https://github.com/googlefonts/gftools/blob/efdc552c7c6578a244cb607ff014ca8a0891e26c/Lib/gftools/subsetmerger.py#L88-L95

### DIFF
--- a/Lib/gftools/subsetmerger.py
+++ b/Lib/gftools/subsetmerger.py
@@ -26,7 +26,7 @@ from ufomerge import merge_ufos
 
 from gftools.gfgithub import GitHubClient
 from gftools.util.styles import STYLE_NAMES
-from gftools.utils import download_file, open_ufo, parse_codepoint
+from gftools.utils import download_file, open_ufo, parse_codepoint, read_glyph_names
 
 logger = logging.getLogger(__name__)
 logging.basicConfig(level=logging.INFO)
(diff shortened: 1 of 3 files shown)


MariannaPaszkowska:
@simoncozens Could you please have a look at this?

simoncozens:
The code LGTM, thank you!

> Side note: there's no practical reason why these two conditionals need to be exclusive, and it means if you include both name and ranges on one subset entry the ranges get silently dropped. Perhaps I could change the elif to an if?

I don't think the semantics of "add a subset which is both a named set of codepoints, and also a range" makes much sense. Just add two subsets...

PR #1177, opened 2026-02-26
builder: add cross-target implicit deps for postprocess steps

m4rc1e:
Zalando Sans has custom recipe targets that call `gftools gen-stat`, https://github.com/zalando/sans/blob/main/sources/config.yaml#L79-L81. Currently, it crashes because `gen-stat` attempts to use fonts that haven't been generated yet.


m4rc1e:
@RosaWagner this will solve the Zalando issue.

PR #1179, opened 2026-03-03
fix-font: add flag to skip rewriting fvar instances

RickyDaMa:
The family I'm migrating to gftools doesn't want to change its instances, but the `fix` operation has other useful goodies so I thought I'd just make this one bit configurable so it would suit our use case.

The ninja template already passes through `$args` so no changes are required in builder land, it just works!

The type hint fairy also came and sprinkled in some goodness ✨

(P.S.: if this could get a release if merged, that would be awesome, thanks!)

### DIFF
--- a/Lib/gftools/fix.py
+++ b/Lib/gftools/fix.py
@@ -878,11 +878,12 @@ def fix_no_varpsname(ttFont: TTFont) -> FixResult:
 
 
 def fix_font(
-    font,
-    include_source_fixes=False,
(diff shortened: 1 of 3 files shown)


m4rc1e:
Thank you for the pr and type hints! LGTM, https://github.com/googlefonts/gftools/releases/tag/v0.9.994

RickyDaMa:
Awesome, thank you Marc!

PR #1180, opened 2026-03-06
Add `gftools-prune` and `prune` builder operation

RickyDaMa:
This is based on a script I see copy-pasted around a lot of repos I maintain. Essentially we use fontTools' subsetter to get rid of dead content of the font without touching anything else. Saves you a few kilobytes here and there, so I thought it would be nice to have upstream. I've not included it in the default recipe or anything drastic, but we have been using this in production for some time, so maybe it's worth considering?

Would appreciate a release with this if merged, as I can replace my custom exec block then :)


Lib/gftools/scripts/prune.py
```diff
@@ -0,0 +1,61 @@
...
+                "--unicodes=*",
+                "--no-ignore-missing-glyphs",
+                "--notdef-outline",
+                "--layout-features=*",
+                "--name-IDs=*",
+                "--name-languages=*",
+                "--glyph-names",
+                "--no-prune-unicode-ranges",
+                "--recalc-bounds",
+                f"--output-file={args.out or args.font}",
```
m4rc1e, line 52: This is the only part that scares me. I'll do some thorough diffs next week to see if there's any flag missing. Other than that LGTM. 

In an ideal world, I'd have this on by default in the builder but I'm too chicken. I may raise it in a pipeline meeting.,

madig:
You could also use hb-subset, like so (untested):

```py
def scrub_font(font_path: Path) -> bytes:
    blob = hb.Blob.from_file_path(font_path)
    face = hb.Face(blob)

    subset_input = hb.SubsetInput()
    subset_input.keep_everything()
    # Don't just keep all glyph indices, let hb use codepoints as a starting point and collect all reachable glyphs:
    subset_input.sets(hb.SubsetInputSets.GLYPH_INDEX).clear()

    cleaned_face = hb.subset(face, subset_input)
    return instance_face.blob.data
```

It's faster.

RickyDaMa:
As far as I know though we haven't used hb-subset in a production environment for font pruning. Whereas the script this PR is based off is used in several product family releases. But yeah, if hb-subset is faster, maybe the eventual goal would be to migrate this operation/script to that. But fontTools still doesn't take long to do this.

RickyDaMa:
Actually, I've just noticed there is actually a `hbsubset` operation which is exactly what I've implemented here, with the option to use either `pyftsubset` (i.e. `fontTools.subset`) or `hb-subset`. I'll close this PR.

https://github.com/googlefonts/gftools/blob/2d2cf4f4e3213b28a2a94962228a989e7909bb48/Lib/gftools/builder/operations/hbsubset.py

PR #1182, opened 2026-03-18
`subsetmerger.py`: use `filelock` to prevent race condition while downloading

### LINKED ISSUE #1181: Race condition between subset downloads
If the recipe for two fonts includes the `addSubset` operation and they both point to the same repo, there is no protection from downloading the zipball twice and then the code getting scared/confused when things already exist that shouldn't etc. as they both download & unpack simultaneously.

Maybe we should use a lockfile or something per repo that's being downloaded to avoid this? I can look into this potentially as I helped with a similar thing with [youseedee](https://github.com/simoncozens/youseedee) in the past

RickyDaMa:
Fixes #1181 

`filelock` is already a transitive dependency, so this isn't adding a new dependency.

I moved the download code path under the original conditional as in the `isinstance(upstream, str) and upstream not in SUBSET_SOURCES` branch, the file's existence has already been checked and the download is never needed, and so we don't need a lockfile. This also fixes "potentially unbound variable" lints for `repo` & `ref`.

### DIFF
--- a/Lib/gftools/subsetmerger.py
+++ b/Lib/gftools/subsetmerger.py
@@ -12,6 +12,7 @@ from zipfile import ZipFile
 
 import ufoLib2
 import yaml
+from filelock import FileLock
 from fontmake.font_project import FontProject
 from fontTools.designspaceLib import (
     DesignSpaceDocument,
@@ -394,6 +395,11 @@ class SubsetMerger:
                 font_name = os.path.basename(upstream)
             else:
                 raise ValueError("Unknown subsetting font %s" % upstream)
+            lockfile_path = os.path.join(
+                upstream,
+                os.path.pardir,
+                f".gftools_subsetmerger_{upstream.replace('/', '_')}.lock",
+            )
         else:
             if isinstance(upstream, str):
                 repo, path = SUBSET_SOURCES[upstream]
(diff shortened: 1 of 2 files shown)


Lib/gftools/subsetmerger.py
```diff
@@ -417,10 +418,15 @@ def obtain_upstream(
                 font_name = f"{repo}/{ref}/{path}"
             path = os.path.join(self.cache_dir, repo, ref, path)
 
-        if os.path.exists(path):
-            logger.info("Subset files present on disk, skipping download")
-        else:
-            self.download_for_subsetting(repo, ref)
+            lockfile_path = os.path.join(
+                self.cache_dir,
+                f".gftools_subsetmerger_{repo.replace('/', '_')}_{ref.replace('/', '_')}.lock",
```
m4rc1e, line 423: LGTM apart from this looks slightly insane at first glance.

RickyDaMa: Yeah that's fair, basically I want the lockfile to be unique to the repo & git ref being used, so the file name would be like `.gftools_subsetmerge_daltonmaag_gftools_lock-for-download.lock` for this branch (though obviously you're using this to download a font source normally). There's some added sanitisation (which is what really dials up the ugly/insane factor) due to the fact that there is guaranteed to be a `/` in the repo location, and there could also be one in the tag name also, and we don't want path separators in our lockfile name

RickyDaMa: I can break it across multiple lines if that'd make it easier on the eyes?

m4rc1e: Nah, life's too short.

PR #1192, opened 2026-06-02
builder: recognise .ufoz (zipped UFO) sources

felipesanches:
`File.is_ufo` only matched an exact `.ufo` extension, so a `.ufoz` (zipped UFO) source was not is_font_source and `File.family_name` crashed with **AttributeError (designspace is None)**. Accept the "ufoz" extension too; `open_ufo()` and fontmake already read `.ufoz`, and the recipe providers / fontc operation pass the source path through unchanged. Behaviour is unchanged for every other extension.

Verified: new unit test (a generated `.ufoz` is a UFO source and resolves a family name) and `gftools builder --generate` on a real `.ufoz` config now emits a correct recipe instead of crashing.

### DIFF
--- a/Lib/gftools/builder/file.py
+++ b/Lib/gftools/builder/file.py
@@ -38,9 +38,7 @@ class File:
     @property
     def is_ufo(self):
         # ".ufoz" is a zipped UFO; ufoLib2 (via open_ufo) reads it like a UFO.
-        return (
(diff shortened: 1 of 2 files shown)


felipesanches:
Note: This is related to https://github.com/googlefonts/fontc/pull/2028

m4rc1e:
Mind running Black on the changed file? Otherwise LGTM.

PR #1195, opened 2026-06-17
Use gfmetadata package instead of protobuf directly

simoncozens:
- **Use gfmetadata for protobuf access**
- **Use gfmetadata for text_format access**
- **diffenator needs protobuf <3, let's just kill it**


m4rc1e:
Thanks!

PR #1198, opened 2026-07-02
Avoid loading Glyphspackage sources with glyphslib where possible

RickyDaMa:
For large projects using the fontc compiler, having to load the font with glyphsLib to check if the font is variable and get the axis tags was a significant proportion of total build time. By just reaching into the fontinfo.plist directly with `openstep_plist`, I reduced a 4 minute build down to a 50 seconds. Admittedly, this font is monstrous, and I'm currently working on a laptop instead of my usual desktop, so the difference is likely to be a lot less prominent for other users. My hope would be that for smaller fonts or projects that do end up needing glyphsLib anyway - thus meaning my approach will have actually added overhead - that this slowdown is negligible.

I understand that most/all of this code is likely to get overwritten by builder 3, but I just wrote it to keep myself entertained while watching CI jobs spin; I don't mind if it's relatively short-lived.

In summary, this PR avoids using glyphsLib to check if a font is variable and read its axis tags, because doing so allows a font compiled with fontc to never need to load the font with glyphsLib, saving time.

### DIFF
--- a/Lib/gftools/builder/file.py
+++ b/Lib/gftools/builder/file.py
@@ -2,6 +2,7 @@ import os
 from dataclasses import dataclass
 from functools import cached_property
 from pathlib import Path
+from typing import Any
 
 import openstep_plist
 from fontTools.designspaceLib import InstanceDescriptor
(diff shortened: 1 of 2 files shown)


simoncozens:
But only for glyphspackages? Could we do the same for .glyphs files too?

RickyDaMa:
Yeah I can do it for both, Harry also suggested it. I initially left it out because with Glyphspackage it means you're reading less data off the disk, but from what I'm hearing most of the load time is in parsing, not reading. I'll make it work on both formats

RickyDaMa:
Done, and checked it worked with a .glyphs version of my favourite monster font ✅

Hoolean:
Just thought of a hitch that I didn't consider earlier, if we look at `.glyphs` we'll have to handle Glyphs 2 data structures too

RickyDaMa:
Tested and support added 🫠 Thanks for flagging that

simoncozens:
You're both brilliant, thank you; this will help dramatically with build speeds.

PR #1206, opened 2026-07-23
Fix gftools-qa

simoncozens:
`gftools-qa` was left in a slightly half-baked state after we committed to Rust tools. The old `--rust` arg was still around in some places, the old `diffenator2.DFont` object was still around in others. This removes both and does some light clean-up to get `gftools-qa` working again.


m4rc1e:
Just had a quick look and there are still some erorrs. I'll fix this today by adding commits to this pr. Thanks for opening.

PR #1211, opened 2026-08-06
[gen-push-lists] Add Tags and Axis registry/Lang sections

emmamarichal:
## Summary

`PushItems.to_server_file()` only writes out categories that exist in the `PushCategory` enum. `tags/all/families.csv` isn't a `PushCategory` member, so it silently fell out of the generated file. The previous workaround (git-diff a check via `pygit2`) only patched this for `to_sandbox.txt`, and only when a diff happened to touch that specific path — `to_production.txt` never got the section at all, and the check was fragile (depended on repo diff state at the time the script ran).

There was also no equivalent reminder for axis registry / lang releases, which are handled manually and easy to forget.

## Change

- Removed the `pygit2`-based conditional check.
- Added `_append_static_footer()`, called unconditionally on **both**
  `to_sandbox.txt` and `to_production.txt` after they're generated. It
  appends:

### DIFF
--- a/Lib/gftools/scripts/gen_push_lists.py
+++ b/Lib/gftools/scripts/gen_push_lists.py
@@ -24,7 +24,6 @@ from gftools.push.utils import branch_matches_google_fonts_main
 from pathlib import Path
 from gftools.utils import is_google_fonts_repo
 from contextlib import contextmanager
-import pygit2
 
 
 @contextmanager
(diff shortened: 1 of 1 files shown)


emmamarichal:
I'm closing this; I thought that with the automation of subtree pulls, the PRs for the Lang and Axis Registry releases would disappear and become hard to track. As it turns out, they're still there, so there's no problem compiling the lists.

PR #1215, opened 2026-08-18
`gftools gen-spac`: support distinct user axis values

RickyDaMa:
For a project, we want to use 200 as the max, but have the axis read as 100. This adds that, by adding two optional flags which overwrite the fvar axis min/max if provided

~~I will work on rebasing #1214 on top of this~~ Update: done

Open to bikeshedding flag names & help strings

### DIFF
--- a/Lib/gftools/scripts/gen_spac.py
+++ b/Lib/gftools/scripts/gen_spac.py
@@ -19,7 +19,14 @@ from fontTools.ttLib.tables.TupleVariation import TupleVariation
 from fontTools.varLib.hvar import add_HVAR
 
 
-def add_spacing_axis(font: TTFont, min_amount: int, max_amount: int) -> None:
+def add_spacing_axis(
+    font: TTFont,
+    min_amount: int,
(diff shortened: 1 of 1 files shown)


m4rc1e:
Can you explain why you need this? I cannot work out why you'd need the SPAC axis min and max to have different vals.

RickyDaMa:
Essentially we want the axis max to be 100 (`--user-max`) - this is to follow the convention outlined in the axis registry ([link](https://github.com/googlefonts/axisregistry/blob/402f39d93efca98724b14735fc5d23ca4bfed95d/Lib/axisregistry/data/spacing.textproto#L6)). We also want the axis to correspond to a ~20% increase in the spacing, which is the `max` value of 200. Maybe it's simpler to show as a designspace snippet:

```xml
<axis tag="SPAC" name="Spacing" minimum="0" maximum="100" default="0">
    <map input="100" output="200"/>
    <map input="0" output="0"/>
</axis>
```

cc @MariannaPaszkowska

PR #1219, opened 2026-08-28
fontc operations: warn on unknown flags/filters (instead of exceptions)

RickyDaMa:
Unknown flags are now passed to fontc verbatim, which will still error if it was a fontmake flag, but otherwise provides a way to give fontc-specific args to fontc when it's being used. Unknown filters are now dropped.

Longer term maybe we should just expect users to use the fontc operations directly (with some extra ease of use in the default recipe provider)? Because bridging this gap is awkward and footgun-prone.

I was also in the locale of an outstanding question I happened to know the answer to, so I wrote that in :)

### DIFF
--- a/Lib/gftools/builder/operations/fontc/__init__.py
+++ b/Lib/gftools/builder/operations/fontc/__init__.py
@@ -1,6 +1,10 @@
+import logging
 from pathlib import Path
 from typing import List
+
 from gftools.builder.operations import OperationBase
+
+log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


anthrotype (approved):

simoncozens:
> Longer term maybe we should just expect users to use the fontc operations directly

Longer term we use gftools-builder3 which has fontc as a first-class citizen and not a somewhat hacky argument-wrapper for fontmake. But sure, we shall do this for the medium term. :-)

PR #1223, opened 2026-09-04
Support deferred source generation in glyphs2ds op

madig:
When switching sources, `glyphs2ds` currently requires the new sources to already exist, which doesn't work when the previous step produced them. Solve this by taking control of the produced Designspace's file path. This also avoids us having to mirror what glyphsLib is doing in some circumstances.


Lib/gftools/builder/__init__.py
```diff
@@ -141,7 +141,8 @@ def config_to_objects(self):
 
     def source_step_to_object(self, step):
         source: str = step["source"]
-        return self._ensure_named_file(source, type="source")
+        target_base_name: str | None = step.get("target_base_name")
```
RickyDaMa, line 144: In the example you showed me, you're doing

```yml
source: foo/MyFont.glyphs
target_base_name: MyFont
```

Why not just use pathlib to get the stem? `Path(step["source"]).stem`

The top-level source stem(s) are already used as a source of truth when constructing the final compiled font file name when using the default googlefonts recipe builder, so it's an established precedent that source names should be sensible. Saves specifying the redundant information with a confusing (imo) attribute name

RickyDaMa (changes requested):

Lib/gftools/builder/__init__.py
madig: Hm, that might be even easier, provided glyphs2ds ensures that the glyphs2ufo op spits out a Designspace file with that stem, need to double check. This would replace the while family name lookup?

Lib/gftools/builder/operations/glyphs2ds.py
```diff
@@ -19,18 +19,13 @@ def targets(self):
...
-        base_family = self.first_source.family_name
-        # glyphsLib has a special case for Glyphs files where the masters have a
-        # common style name; typically this manifests itself in single-master
-        # Glyphs files, but it can also happen in multi-master files.
-        # In that case, the designspace filename will be changed.
-        if base_style := find_base_style(self.first_source.gsfont.masters):
-            base_family += "-" + base_style
-        dspath = os.path.join(target, base_family.replace(" ", "") + ".designspace")
+        base_name = Path(self.first_source.path).with_suffix(".designspace").name
+        dspath = os.path.join(target, base_name)
```
RickyDaMa, line 23: I'd add a comment here that this relies on the sources being well-named, which is an understood convention elsewhere in the codebase.

I'm not very familiar with what the old logic is doing, so a test case that exercises that special case would be interesting to see

RickyDaMa (commented):
I like the principle of the changes, I just don't understand the bigger picture enough to feel confident about breaking changes / regressions. Unit tests or just anecdotal testing would be reassuring.

Lib/gftools/builder/operations/glyphs2ds.py
madig: What do you mean by well-named?

RickyDaMa: That the newly created source has a useful name for the final font. If I'm correctly remembering how gftools works, the font compiled by `glyphs2ds` in your test will be called `frobnicated.ttf`, named after its source. So the name given to the source is important because it's used for the font file name, which I think should be documented. If that's not how it works then feel free to ignore me, clearly my memory is wrong/outdated (maybe it only works like that if `familyName` is not specified?).

madig: So, the end result coming out of gftools config.yaml with a familyName is a correctly named font, but I'm not sure if I'm overlooking something.

simoncozens:
I'm happy with this but it looks like I can't push to your repo; run black, please?

Lib/gftools/builder/operations/glyphs2ds.py
RickyDaMa: What if you don't set `familyName`? gftools has a fallback I'm fairly certain

madig: Then it still comes out fine... Maybe it's taking the defined file name attached to the recipe?

PR #1224, opened 2026-09-11
packager: prevent path traversal via untrusted METADATA source.files …

ghost:
## Summary

`gftools packager` uses contributor-controlled `source.files[].dest_file`
values from a family's metadata to construct output paths.

Those paths were joined directly to the package output directory without
constraining the resolved result to that directory. An absolute path or
`..` traversal could therefore cause the packager to write downloaded asset
contents outside the intended family directory.

This crosses the trust boundary from contributor-controlled metadata to the
maintainer's local filesystem when the packager is used to process a family.

## Fix

Route all `dest_file` joins through a root-constrained path helper that:

- rejects absolute paths
- resolves the candidate path
- rejects results outside the intended output root

All three identified `dest_file` sinks are covered.

## Tests

Added regression coverage for:

- normal relative paths
- nested relative paths
- `..` traversal
- absolute paths
- mixed traversal

The local PoC reproduced the traversal before the fix and confirmed that
the same inputs are blocked after the fix.

Depending on the destination selected by an attacker, arbitrary file write
could also enable code execution when the written location is subsequently
executed by the maintainer's environment.

### DIFF
--- a/Lib/gftools/packager/__init__.py
+++ b/Lib/gftools/packager/__init__.py
@@ -27,6 +27,7 @@ from pygit2 import (
 from pygit2.enums import FileStatus
 
 from gfmetadata import FamilyProto, SourceFileProto
+from gftools.packager._pathsafe import safe_join
 from gftools.gfgithub import GitHubClient
 from gftools.scripts.add_font import main as add_font
 from gftools.util import google_fonts as fonts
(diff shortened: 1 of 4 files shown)


simoncozens:
Thank you for your PR! I notice that in the past three days you have created PRs in google/nsjail, tscircuit/circuit-json-util, protobuf, tscircuit/contribution-tracker and now gftools. You are either someone of diverse interests, or an AI bot looking for security bounties.

```
``METADATA.pb`` / ``upstream.yaml`` ``source.files[].dest_file`` values are
contributor-controlled (they arrive with the family being on-boarded). 
```

I don't think that's true; METADATA.pb is written by the person running `gftools-packager`.

m4rc1e:
As Simon said the tooling itself makes these files so I trust it.

